PR #8, opened 2021-01-07
Rename card-rulesets to card-workflows

cbetta:
Renamed `card-rulesets` to `card-workflows`. This prevents confusion between `rulesets` and `rules`, and also is a more generally understandable term.


danielknell (approved):
LGTM

PR #10, opened 2021-01-08
Payment method options and pagination optimization

cbetta:
* Remove pagination from payment options
* Remove supported currencies/countries from payment options
* Introduce a short-format (`token` format) of a payment method
* Update payment method API to allow for selecting format
* Add list-buyer-payment-methods API to return token-format payment methods for a buyer


douglaseggleton:
- Retrieving the list of payment options - Does this include both "token" resources and "full" resources? or will this be a separate endpoint I would need to call?

- Payment Method, e.g. Card, Paypal  -  Payment Option, e.g. Tokenized Card, New Card ... is this correct terminology?

reference/resources/card--token.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+  token:
+    type: string
+    description: |-
+      A Gr4vy token that represents this payment method, in the format of a
+      credit-card number starting with the digit 0. You can store this token
+      in your own system and use it to process a payment with this payment
+      method at a later date.
+    example: "0123456789123458"
+    nullable: true
```
douglaseggleton, line 24: Does nullable man a Card Token resource can exist without a token?

douglaseggleton (approved):

aliminaei (approved):
Looks good to me

reference/resources/card--token.yaml
cbetta: Yes. But thos only affects the list-payment-methods API. In the list-buyer-payment-methods API we will only return tokenized payment methods.

PR #12, opened 2021-01-15
refactor(approvals): Refactor approvals to be a state of a transaction

cbetta:
- Rename payment method approval/confirm endpoints
- Add transaction approval-redirect/approve endpoints
- Add new approved status to payment methods and transactions
- Add new endpoint to authorize a previously approved resource
- Update payment method/transaction status enums to include new approved status
- Disable tokenizing paypal payment methods (for now)
- Refactor the `capture` enum into an `intent` with possible values `approve`, `authorize`, `capture`.

### DIFF
--- a/reference/openapi.yaml
+++ b/reference/openapi.yaml
@@ -137,6 +137,12 @@ paths:
     post:
       $ref: "./paths/transactions__async-create-transaction.yaml"
 
+  "/async/transactions/{transaction_id}/authorize":
+    parameters:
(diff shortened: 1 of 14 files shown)


aliminaei:
Nothing to do with this change but the only thing I noticed is the CI/CD path, we have been using `ci_cd` in our other repos

aliminaei (approved):

cbetta:
@aliminaei CI path needs to be `workflows` cause that's where GitHub Actions expect them

PR #14, opened 2021-01-22
Re-introduce buyer-payment-methods endpoint 

cbetta:
Adding the `GET /buyers/payment-methods` API. It returns a different shorter format for use in Embed and allows for searching of payment methods by buyer ID. It also allows for filtering by country and currency as well.


cbetta:
Merged into #17

PR #17, opened 2021-01-27
Buyer and payment method updated

cbetta:
- [x] Add buyer search
- [x] Add buyer-payment methods endpoint
- [x] Remove token from tokenized payment methods
- [x] Updated tokenized payments to use the payment method ID
- [x] Dropped the format query for the `GET /payment-methods` endpoint


danielknell (approved):

douglaseggleton (approved):
LGTM. I assume a Card is a Card, there's nothing different per say about a tokenized resource apart from the status being "stored" and the additional data that will come back with the resource. e.g. obfuscated card number

cbetta:
> LGTM. I assume a Card is a Card, there's nothing different per say about a tokenized resource apart from the status being "stored" and the additional data that will come back with the resource. e.g. obfuscated card number

Yeah a card resource is the same wether it's tokenized or not, just a different status.

PR #18, opened 2021-01-27
Rework environments and simplify definitions

cbetta:
Based on https://docs.google.com/document/d/1MQukTJdHfhWUgU82G9l_DP2uUGgbNF_xtrLHVIR97LE/edit#

- [x] Replaced `use_test_environment` with a few new fields: `active`, `credentials_mode` and `environments`.
- [x] Added the ability to specify the intended `environment` across the payment method and transaction APIs.
   - [x] Additionally add the `environments` flag to buyer payment methods. 
- [x]  Vastly simplified the payment service definitions, dropping all `help_url` and `logo_url` fields. Many of these field weren't used as I realised we really need a guide for every payment service that shows how to set up that provider with Gr4vy. https://github.com/gr4vy/gr4vy-api-services/pull/71


alibeylan (dismissed):
LGTM (:

PR #22, opened 2021-02-09
refactor(rules): Remove boolean conditionals for rules

cbetta:
Removing because:

1. Linking rules to environments should be part of the rule, not a conditional
2. Acting upon if a card was being stored had no valid usecase.


aliminaei (approved):
LGTM

PR #40, opened 2021-06-07
Add payment option locale parameter/label property

douglaseggleton:
# Description

Currently in development - This will be the API definition for translated labels on payment options.

Screenshot 2021-06-07 at 12 47 31

## Checklist

## Contribution guidelines

### DIFF
--- a/reference/query-parameters/locale.yaml
+++ b/reference/query-parameters/locale.yaml
@@ -7,4 +7,5 @@ schema:
 in: query
 name: locale
 description: |-
-  An ISO 639-1 Language Code and optional ISO 3166 Country Code to translate messages.
+  An ISO 639-1 Language Code and optional ISO 3166 Country Code to translate
+  messages.
--- a/reference/query-parameters/locale.yaml
+++ b/reference/query-parameters/locale.yaml
@@ -1,0 +1,8 @@
+---
+schema:
+  type: string
+  example: en-US
+in: query
+name: locale
+description: |-
+  An ISO 639-1 Language Code and optional ISO 3166 Country Code to translate messages.
--- a/reference/query-parameters/locale.yaml
+++ b/reference/query-parameters/locale.yaml
@@ -7,5 +7,5 @@ schema:
 in: query
 name: locale
 description: |-
-  An ISO 639-1 Language Code and optional ISO 3166 Country Code to translate
-  messages.
+  An ISO 639-1 Language Code and optional ISO 3166 Country Code. This locale determines
+  the language for the labels returned for every payment option.
--- a/reference/query-parameters/locale.yaml
+++ b/reference/query-parameters/locale.yaml
@@ -2,6 +2,8 @@
 schema:
   type: string
   example: en-US
+  pattern: ^[a-z]{2}(?:-[A-Z]{2})?$
+  default: en-US
 in: query
 name: locale
 description: |-
--- a/reference/resources/payment_option.yaml
+++ b/reference/resources/payment_option.yaml
@@ -35,3 +35,9 @@ properties:
     enum:
       - card
       - redirect
+  
+  label:
+    type: string
+    description: The label to show for the the payment option in the requested locale
+    example: Pay by Card
+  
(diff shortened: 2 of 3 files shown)


reference/query-parameters/locale.yaml
```diff
@@ -0,0 +1,8 @@
+---
+schema:
```
danielknell, line 2: does this need pattern along the lines of `pattern: ^[a-z]{2}(?:-[A-Z]{2})?$` field to specify the format?

gryevns:
Should we look at using the `Accept-Language` header rather than add an optional query param to all endpoints?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language

reference/resources/payment_option.yaml
```diff
@@ -35,3 +35,9 @@ properties:
     enum:
       - card
       - redirect
+  
+  label:
+    type: string
+    description: The label to show for the the payment option in the requested locale
+    example: Pay by Card
```
danielknell, line 42: i still have questions around this that i think @cbetta will need to look at, will these always be "Pay by X" meaning we can just return X, or will they vary from one service to another? only example we could think of earlier was a "Pay later with X" style wording for the payment plan services...

the name feels like its more useful outside of embed.

danielknell:
> Should we look at using the `Accept-Language` header rather than add an optional query param to all endpoints?
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language

we looked at that, but its possibly overkill to support a prioritised list of locals and would be weird if that worked in some places but not others...

douglaseggleton:
> > Should we look at using the `Accept-Language` header rather than add an optional query param to all endpoints?
> > https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language
> 
> we looked at that, but its possibly overkill to support a prioritised list of locals and would be weird if that worked in some places but not others...

We just talked about this :D It specifically just this endpoint, so we felt it would be simpler to keep it as a param.

reference/resources/payment_option.yaml
douglaseggleton: How would you solve something like this?

Betala per telefon.  🇸🇪
Betala med kort  🇸🇪

I don't speak many languages - but it may not make "sense" as phrase.

reference/query-parameters/locale.yaml
```diff
@@ -0,0 +1,11 @@
...
+  type: string
+  example: en-US
+  pattern: ^[a-z]{2}(?:-[A-Z]{2})?$
+  default: en-US
+in: query
+name: locale
+description: |-
+  An ISO 639-1 Language Code and optional ISO 3166 Country Code to translate
+  messages.
```
cbetta, line 11:
```suggestion
  An ISO 639-1 Language Code and optional ISO 3166 Country Code. This locale determines
  the language for the labels returned for every payment option.
```

reference/resources/payment_option.yaml
```diff
@@ -35,3 +35,10 @@ properties:
     enum:
       - card
       - redirect
+
+  label:
+    type: string
+    description: |-
+      The label to show for the the payment option in the requested
+      locale.
```
cbetta, line 43:
```suggestion
      A label that describes this payment option. This label is returned in the language 
      defined by the `locale` query parameter. The label can be used to display a list
      of payment options to the buyer in their language.
```

reference/resources/payment_option.yaml
cbetta: Definitely should not be a header as that would put the decision on the browser and not Embed.

cbetta (approved):
A few small changes requested. Looks good beyond that though.

reference/resources/payment_option.yaml
danielknell: that was a different discussion though...?

danielknell: considering i speak exactly 0.5 languages, i guess ill defer to doug on the point above, though it makes this very embed specifc vs other possible integrations.

PR #55, opened 2021-08-03
feat: adding hosted 3-D Secure merchant data

douglaseggleton:
Updated to reflect the API changes for the configuring 3-D Secure
https://sprints.zoho.com/team/gr4vy#itemdetails/P10/I257


reference/resources/payment_service_definition.yaml
```diff
@@ -69,6 +69,39 @@ properties:
             - multiline
             - number
 
+  three_d_secure_fields:
+    type: array
```
gryevns, line 73: I was expecting this to be an object rather than an array (there will only ever be one set of 3DS data)

reference/resources/payment_service.yaml
```diff
@@ -83,6 +83,48 @@ properties:
...
       - DE
 
+  three_d_secure_enabled:
+    type: boolean
+    example: true
+    default: false
+    description: |-
+      Defines if 3-D Secure is enabled for the service (can only be
+      enabled if the payment service definition supports
+      the `three_d_secure_hosted` feature).
```
gryevns, line 93: This should be part of the `three_d_secure_details` object to keep all 3DS data together?

reference/resources/payment_service_definition.yaml
douglaseggleton: I went away and thought about this. 🤔

Why do you prefer it to be an object?
From my perspective, the **fields should be ordered** (array) and this follows the same pattern as `fields`.

I then went further, as this concerns grouped fields - we can define ordering/groups etc. separately from field definitions, but also the dashboard doesn't need to know anything about the specifics of 3D-Secure.

```ts
fields: {
  merchant_id: {
    label: 'Merchant ID', // etc
  }
  // map of fields defs.
}

fieldsets: [{
  name: 'Details',
  fields: ['name', 'api_key'] // reference field definitions
}, {
  name: '3-D Secure',
  fields: ['merchant_id', 'merchant_name']
}]

```

gryevns: Oh I misread, it's an array of objects where each object represents a different input - that makes sense.

reference/request-bodies/payment_service_update_request.yaml
```diff
@@ -73,6 +73,47 @@ properties:
...
       - GBP
 
+  three_d_secure_enabled:
+    type: boolean
+    example: true
+    default: false
+    description: |-
+      Defines if 3-D Secure is enabled for the service (can only be
+      enabled if the payment service definition supports
+      the `three_d_secure_hosted` feature).
```
ipeluffo, line 83: Should we add a clarification that even when this is disabled 3DS pass through will be used if the data was sent? 🤔

reference/resources/payment_service.yaml
ipeluffo: Does this mean that `three_d_secure_details` will never be `null` and contain all keys?

reference/request-bodies/payment_service_update_request.yaml
```diff
@@ -73,6 +73,47 @@ properties:
       - USD
       - GBP
 
+  three_d_secure_enabled:
```
ipeluffo, line 76: if we move this into `three_d_secure_details` (as per [comment](https://github.com/gr4vy/gr4vy-openapi/pull/55/files#r681894348)), then the update probably should send the full details on each update. I'm not sure if this impacts the documentation.

reference/resources/payment_service_definition.yaml
```diff
@@ -69,6 +69,39 @@ properties:
             - multiline
             - number
 
+  three_d_secure_fields:
```
ipeluffo, line 72: Do we really need this as it'll be the same for all payment services? I thought we were going to hardcode them in dashboard side (maybe I'm wrong 😅 )

reference/request-bodies/payment_service_update_request.yaml
douglaseggleton: We should probably call that out in the UI too 👀

ipeluffo: hmm... not sure because we're not going to send 3DS data (from Gr4vy) if 3DS is not enabled for the payment service used to process the transaction (hopefully that makes sense)

ipeluffo: > We should probably call that out in the UI too 👀

Do you mean in Dashboard?

douglaseggleton: Yep - Dashboard 💯

PR #65, opened 2021-09-23
docs: add 3-D Secure api changes

douglaseggleton:
# Description
Adding the API changes for 3-D Secure flows (both external and redirected/hosted)

## Changes to POST /transactions (Payment Method)
I have put the redirect url as part of the payment method, its actually required if three_d_secure_data is not provided.
I don't know what the best way to do this in Open API.

Screenshot 2021-09-23 at 15 11 51

## Changes to POST /transactions (Three D Secure Data)
I followed the backend rules from the code, but you might need to check there - e.g. I assume maxLength 1 also means minLength 1?

Screenshot 2021-09-23 at 15 11 13
Screenshot 2021-09-23 at 15 11 06

# Fixes
https://sprints.zoho.com/team/gr4vy#itemdetails/P10/I437


gryevns (commented):
Looks good, worthwhile fixing the linting warnings for the new changes?

gryevns (approved):

PR #90, opened 2021-12-10
feat: Add statement descriptor for new transaction endpoint [TA-559]

ipeluffo:
# Description

This PR adds `statement_descriptor` in "New Transaction" endpoint.

Fixes # (issue)
https://gr4vy.atlassian.net/browse/TA-559

## Checklist

## Contribution guidelines


reference/resources/statement_descriptor.yaml
```diff
@@ -0,0 +1,75 @@
+---
+title: Statement descriptor
```
ipeluffo, line 2: @cbetta @steve-gr4vy 

This PR is not ready yet because I'm waiting for changes on another PR but it'd be great to have some feedback from you about the validations and descriptions for the statement descriptor data. The PR won't change much so your feedback probably will be most of the review for this PR.

Thanks in advance

cbetta (approved):

bloowitt (approved):

reference/resources/statement_descriptor.yaml
```diff
@@ -0,0 +1,77 @@
...
+    minLength: 5
+    maxLength: 20
+    description: The value in the phone number field of a customer's statement.
+    example: "+1234567890"
+    nullable: true
+
+  url:
+    type: string
+    minLength: 1
+    maxLength: 13
```
bloowitt, line 73: shouldn't this be maxLength=50 at least? 13 seems short for a url

reference/resources/statement_descriptor.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+  city:
+    type: string
+    minLength: 1
+    maxLength: 13
+    description: City from which the charge originated.
+    example: London
+    nullable: true
+
+  phone:
```
bloowitt, line 62: Sorry to be late with this, but I just realised: in Buyers we are already using 'phone_number' with a max length of 50 for phone numbers. Should we adopt both here for homogeneity on the API?

bloowitt (commented):
Left a question I think it's worth thinking about before merging this

reference/resources/statement_descriptor.yaml
cbetta: I think there's a limit as to what these fields support in the PSPs. They're not the same phone/URL fields as we use in other places. I'll let @ipeluffo answer.

cbetta (approved):

reference/resources/statement_descriptor.yaml
ipeluffo: I set this length based on [Braintree docs](https://developer.paypal.com/braintree/docs/reference/request/transaction/sale#descriptor.url) as it was the only PSP from the ones I checked that supports URL field.

That said, we can allow a longer value and mention in the description that the value can be concatenated. So, 50 sounds sensible to me.

@cbetta , any thoughts?

reference/resources/statement_descriptor.yaml
ipeluffo: Yeah, similar answer as for URL. I took this from [Braintree docs](https://developer.paypal.com/braintree/docs/reference/request/transaction/sale#descriptor.phone) that says:

> The value in the phone number field of a customer's statement. Phone must be 10-14 characters and can only contain numbers, dashes, parentheses and periods.

From what I've read, descriptors are weird as they are usually not validated and "wrong" values are set in fields like an URL in a city field. Because we have multiple fields for a descriptor, maybe we can validate values.

ipeluffo: Also.... should we rename this as `phone_number` to match the name in billing details?

bloowitt: Yeah, happy with the size limit if it's based on underlying PSPs. Still think renaming it would be good practise (and maybe start an style guide for the API so we know these things when first writing it?

ipeluffo: I'm happy with renaming but we'll need to propagate this change in the spec and implementation for connectors

bloowitt (dismissed):

reference/resources/statement_descriptor.yaml
cbetta: I think this isn't really a URL. It can really be any string, and it's only used by Braintree. So let's stay close to their spec.

cbetta (approved):

PR #93, opened 2022-01-14
feat(digital wallets): session api docs

gryevns:
# Description

Adds Google Pay session endpoint docs and consolidates Apple Pay session endpoint.

Requires: https://github.com/gr4vy/core-api/pull/524

## Checklist

## Contribution guidelines


reference/paths/digital_wallets_apple_pay__session.yaml
```diff
@@ -0,0 +1,17 @@
+---
+operationId: new-apple-pay-session
+summary: New Apple Pay session
+description: Initiates a new session with Apple Pay.
+tags:
+  - Apple Pay
```
szeying-lee, line 6: Should this be under the `Digital Wallets` tag? Or are we keeping them separate because it's internal only?

reference/paths/digital_wallets_google_pay__session.yaml
```diff
@@ -0,0 +1,17 @@
+---
+operationId: new-google-pay-session
+summary: New Google Pay session
+description: Initiates a new session with Google Pay.
+tags:
+  - Google Pay
```
szeying-lee, line 6: Ditto above, but if we want to keep this under its own tag then I think `Google Pay` should also be defined here: 
https://github.com/gr4vy/gr4vy-openapi/blob/main/reference/openapi.yaml#L28

reference/paths/digital_wallets_apple_pay__session.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+requestBody:
+  content:
+    application/json:
+      schema:
+        $ref: "../openapi.yaml#/components/schemas/ApplePaySessionRequest"
+
+responses:
+  "200":
+    description: Returns an opaque Apple Pay session object.
```
szeying-lee, line 17: Should there be 4xx responses like in the original file? Same question for `digital_wallets_google_pay__session.yaml`.

szeying-lee (approved):

PR #98, opened 2022-01-25
feat: TA-92 - Add new refund/void transaction endpoints

federicojasson:
# Description

This PR introduces the new refund/void transaction endpoints, while puts a deprecation warning on the old one.

## Checklist

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own changes
- [X] I have run `yarn test` to make sure my changes pass all linters
- [X] I have pulled the latest changes from the upstream `main` branch

## Contribution guidelines


federicojasson:
This PR should probably be merged only after the feature is implemented.

reference/paths/transactions__refund-transaction.yaml
```diff
@@ -1,12 +1,14 @@
...
-summary: Refund or void transactions
-
+summary: Refund transaction
 description: |-
-  Refunds or voids transaction. If this transaction was already captured, it
-  will issue a refund. If the transaction was not yet captured the authorization
-  will instead be voided.
+  Refunds a transaction, fully or partially.
+
+  If the transaction was not yet captured, the behaviour depends on the
```
bloowitt, line 8: According to TA-91:
```
The existing refund endpoint in the connector will need to be split into separate void and refund endpoints. Each will return a failed response when the action is not applicable:

    void will fail if the transaction is processing or already captured

    refund will fail if the transaction has not yet been captured
```
  So it should not reach the PSP if the status is not appropriate. Something like "the transaction needs to have been captured for this endpoint to work`?

bloowitt (changes requested):
Left a comment on endpoint behaviour

reference/paths/transactions__refund-transaction.yaml
cbetta: Yes. 100% this. This new endpoint will only process the refund if the transaction status is `capture_succeeded` (check status in core API to confirm spelling)

reference/paths/transactions__refund-transaction.yaml
```diff
@@ -1,12 +1,14 @@
...
+  underlying PSP. In that case, use the
+  [Void transaction](#operation/void-transaction) endpoint instead.
```
cbetta, line 10:
```suggestion
  If the transaction was not yet successfully captured the refund will not be processed.
```

reference/paths/transactions__refund-transaction.yaml
```diff
@@ -15,14 +17,32 @@ requestBody:
     application/json:
       schema:
         $ref: "../openapi.yaml#/components/schemas/TransactionRefundRequest"
+      examples:
+        Request a full refund on a transaction:
+          value: {}
```
cbetta, line 22: Do we want to support an empty request body?

reference/paths/transactions__void-transaction.yaml
```diff
@@ -0,0 +1,56 @@
+---
+operationId: void-transaction
+
+summary: Void transaction
+description: |-
+  Voids a transaction.
+
+  If the transaction was already captured, the behaviour depends on the
+  underlying PSP. In that case, use the
+  [Refund transaction](#operation/refund-transaction) endpoint instead.
```
cbetta, line 10:
```suggestion
  If the transaction was not yet successfully authorized, or already captured the void
  will not be processed. Captured transactions can be [refunded](#operation/refund-transaction)
  instead.
```

reference/request-bodies/transaction_refund_request.yaml
```diff
@@ -10,9 +11,11 @@ properties:
   amount:
     type: integer
     description: |-
-      The (partial) amount to refund.
+      The amount requested for refund.
```
cbetta, line 14:
```suggestion
      The amount requested to refund.
```

reference/request-bodies/transaction_refund_request.yaml
```diff
@@ -10,9 +11,11 @@ properties:
...
 
-      When omitted blank, this will refund the entire amount.
-    example: 1299
+      If omitted, a full refund will be requested.
+      Otherwise, the amount must be lower than or equal to the remaining balance
+      in the associated transaction.
```
cbetta, line 18:
```suggestion
      If omitted, a full refund will be requested.
      Otherwise, the amount must be lower than or equal to the remaining balance
      in the associated transaction. Negative refunds are not supported.
```

reference/resources/refund.yaml
```diff
@@ -0,0 +1,75 @@
...
+    description: |-
+      The status of the refund.
+      It may change over time as asynchronous processing events occur.
+
+      - `processing` - The refund is being processed.
+      - `succeeded` - The refund was successful.
+      - `declined` - The refund was declined by the underlying PSP.
+      - `failed` - The refund could not proceed due to a technical issue.
+      - `voided` - The refund was voided and will not proceed.
+    enum:
```
cbetta, line 38: Are these states based on @danielknell's initial list?

reference/resources/refund.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+  updated_at:
+    type: string
+    description: The date and time when this refund was last updated.
+    format: date-time
+    example: "2013-07-16T19:23:00.000+00:00"
+
+  xid:
+    type: string
+    description: The transaction identifier in the underlying PSP.
```
cbetta, line 74:
```suggestion
    description: The refund identifier in the underlying PSP.
```

reference/resources/refund.yaml
```diff
@@ -0,0 +1,75 @@
...
+x-tags:
+  - Transactions
+
+properties:
+  id:
+    type: string
+    description: The unique ID of the refund.
+    format: uuid
+    example: 8724fd24-5489-4a5d-90fd-0604df7d3b83
+
```
cbetta, line 20: Every resource at Gr4vy requires an `id` and `type`. Type here should be `refund`.

reference/resources/refund.yaml
```diff
@@ -0,0 +1,75 @@
...
+  updated_at:
+    type: string
+    description: The date and time when this refund was last updated.
+    format: date-time
+    example: "2013-07-16T19:23:00.000+00:00"
+
+  xid:
+    type: string
+    description: The transaction identifier in the underlying PSP.
+    maxLength: 300
```
cbetta, line 75: Do we return the `xid` anywhere else in our API? I don't think we do. Please leave it out, as I think the term `xid` is purely internal atm.

reference/paths/transactions__void-transaction.yaml
```diff
@@ -0,0 +1,56 @@
...
+  - Transactions
+
+requestBody:
+  content:
+    application/json:
+      schema:
+        $ref: "../openapi.yaml#/components/schemas/TransactionVoidRequest"
+      examples:
+        Request to void a transaction:
+          value: {}
```
cbetta, line 22: I think we can leave this whole request body out as it's empty. It's basically a `POST` without a body.

cbetta (changes requested):
Requested a few changes to the copy, as well as some structural ones.

reference/resources/refund.yaml
federicojasson: Based on https://gr4vy.atlassian.net/wiki/spaces/GB/pages/153125577/Refund+model

reference/paths/transactions__refund-transaction.yaml
federicojasson: Well, it's the default state of the object when all optional fields (in this case, just `amount`) are left out. Also, I'm not sure how to display the example more clearly than with an empty object, which contrasts with the partial refund example.

reference/resources/refund.yaml
danielknell: the only place we expose xid currently is the `tokens` endpoint which renames it to `id`

bloowitt (dismissed):
LGTM now, I'll leave it in @cbetta's hands

epigos (dismissed):
LGTM

reference/paths/transactions__refund-transaction.yaml
cbetta: Yeah fine as is.

cbetta (approved):
LGTM!

bloowitt (approved):

PR #110, opened 2022-03-03
feat: add ip_address [TA-809]

szeying-lee:
# Description
Adds `ip_address` as a new field to all `/transactions` and `/payment-methods` endpoints that make a call to a connector.

Fixes # (issue)
https://gr4vy.atlassian.net/browse/TA-809

## Checklist

## Contribution guidelines


reference/request-bodies/card_request.yaml
```diff
@@ -73,3 +73,8 @@ properties:
       to. If this field is provided then the `buyer_id` field
       needs to be unset.
     example: user-789123
+
+  ip_address:
```
thomas-gr4vy, line 77: `ip_address` is also on the transaction create schema. What will happen if I provide an IP address in both?

thomas-gr4vy (commented):
Looks good. Just one question RE the transaction create endpoint.

reference/request-bodies/card_request.yaml
szeying-lee: The POST /transactions logic for this change in Core API only uses the outer `ip_address`, meaning any `ip_address` passed as part of the payment method is ignored but won't raise any errors. Do you think I should be checking for both in case `ip_address` is supplied in the payment method but not in the outer transaction? 🤔

reference/paths/payment_methods__approve-payment-method.yaml
```diff
@@ -17,6 +17,9 @@ tags:
 x-internal: true
 security: []
 
+parameters:
+  - $ref: "../query-parameters/ip_address.yaml"
+
```
cbetta, line 22: I guess we need to do this cause we don't intend to store the IP?

reference/paths/payment_methods__store-payment-method.yaml
```diff
@@ -11,8 +11,22 @@ requestBody:
...
-          - $ref: ../openapi.yaml#/components/schemas/CardRequest
-          - $ref: ../openapi.yaml#/components/schemas/RedirectRequest
+          - allOf:
+              - $ref: ../openapi.yaml#/components/schemas/CardRequest
+              - type: object
+                properties:
+                  ip_address:
+                    type: string
+                    description: The IP address of the buyer.
+                    example: "111.111.111.111"
```
cbetta, line 21: Why not just add this to `CardRequest` and `RedirectRequest`?

cbetta: Definitely not the right way to do this (sorry). 

1. Properties should not be listed here, the new properties should be part of a named schema (e.g. CardRequest)
2. We should make the value nullable
3. We should maybe just design the API to expect the IP always within the same place, regardless of API call.

Happy to jump on a call to do this the right way together.

cbetta (changes requested):

reference/paths/payment_methods__approve-payment-method.yaml
szeying-lee: This particular endpoint is a GET endpoint even though it sends a request to the connector. Although FastAPI supports sending a request body with a GET, it isn't recommended: https://fastapi.tiangolo.com/tutorial/body/

reference/paths/payment_methods__store-payment-method.yaml
szeying-lee: My initial changes did add them to `CardRequest` and `RedirectRequest`, however doing so would allow `ip_address` to be sent as part of these payment methods in create transaction requests as well. See also the point Thomas raised: https://github.com/gr4vy/gr4vy-openapi/pull/110#discussion_r824554601

Have discussed with him and decided it would be best that the schemas should ask only for what is needed, so removed it out of those resources and extended it this way. However, I see your point and I should have made a new schema, maybe something like `CardRequestWithIpAddress` or similar, that extends `CardRequest` and adds `ip_address` as an extra property, and then reference it here.

cbetta:
Closing this in favour of a different approach after conversation with @szeying-lee

PR #117, opened 2022-03-21
feat: Add /buyers/billing_details endpoints [TA-1022]

szeying-lee:
# Description
Documents new `PUT /buyers/billing_details` and `GET /buyers/billing_details` endpoints. Also recreates `address_update_request.yaml` as all address fields are actually optional (https://github.com/gr4vy/core-api/blob/main/src/api/routers/buyers/schemas.py#L91).

Fixes # (issue)
https://gr4vy.atlassian.net/browse/TA-1022

## Checklist

## Contribution guidelines


reference/openapi.yaml
```diff
@@ -187,6 +187,12 @@ paths:
     delete:
       $ref: "./paths/buyers__delete-buyer.yaml"
 
+  "/buyers/billing_details":
```
thomas-gr4vy, line 190:
```suggestion
  "/buyers/billing-details":
```

reference/paths/buyers__get-buyer-billing-details.yaml
```diff
@@ -0,0 +1,44 @@
...
+  - $ref: "../query-parameters/buyer_id.yaml"
+  - $ref: "../query-parameters/buyer_external_identifier.yaml"
+
+responses:
+  "200":
+    description: Returns the information about a buyer's billing details.
+    content:
+      application/json:
+        schema:
+          $ref: "../openapi.yaml#/components/schemas/BillingDetails"
```
thomas-gr4vy, line 23: Do we need to extend this to add a `type` of `billing-details` (and what about an `id`)?

thomas-gr4vy (commented):
Added a few comments.

reference/request-bodies/address_update_request.yaml
```diff
@@ -5,15 +5,81 @@ type: object
...
-  - line1
-  - line2
-  - city
-  - state
-  - postal_code
-  - country
-
-allOf:
-  - $ref: "../openapi.yaml#/components/schemas/Address"
+properties:
```
cbetta, line 8: This schema is also used in other requests? Is this changing the behaviour of these? E.g. is it setting fields as not required in them?

szeying-lee: Yes, `AddressUpdate` is used in `BillingDetailsUpdateRequest` which is used in `BuyerUpdate`. Have added a link in the description describing this change as none of the fields in `AddressUpdate` is actually required: https://github.com/gr4vy/core-api/blob/main/src/api/routers/buyers/schemas.py#L91

cbetta: @szeying-lee perfect.

reference/paths/buyers__get-buyer-billing-details.yaml
cbetta: Agreed with this. Not sure what the ID would be though.

cbetta: FWIW we could add the type to all BillingDetails records, even those nested in buyers. The ID could be a unique UUID. Not sure if we have a UUID already internally?

thomas-gr4vy (approved):

reference/resources/billing_details.yaml
```diff
@@ -1,18 +1,25 @@
...
   - Buyers
 
 properties:
+  type:
+    type: string
+    description: The type of this resource. Is always `billing-details`.
+    example: billing-details
+    enum:
+      - billing-details
+
```
cbetta, line 17: Does this still need an ID? @thomas-gr4vy ?

thomas-gr4vy: Thinking about it, I'm not sure it makes sense to add it at this point in time. There is no way to use the ID to access the resource on its own. Maybe one for the future if/when we support multiple billing details per buyer?

PR #128, opened 2022-04-05
feat: TA-1135 - Document void/refund and verify_credentials feature flags

federicojasson:
# Description

This PR documents the new void/refund feature flags.

Additionally, it documents the previously-added feature flag `verify_credentials` (ticket [TA-879](https://gr4vy.atlassian.net/browse/TA-879)).

Ticket: [TA-1135](https://gr4vy.atlassian.net/browse/TA-1135)

## Checklist

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own changes
- [X] I have run `yarn test` to make sure my changes pass all linters
- [X] I have pulled the latest changes from the upstream `main` branch


reference/resources/payment_service_definition.yaml
```diff
@@ -120,6 +120,30 @@ properties:
         description: |-
           Supports passing decrypted digital wallet (e.g. Apple Pay)
           tokens to the underlying processor.
+      verify_credentials:
```
federicojasson, line 123: I'm doubtful about adding this feature flag or not. The ticket specifies that it should have been added, but I understand this is a private operation.

cbetta: If it's visible in a public URL then we should document it. Maybe say it's for internal use only?

federicojasson: Feature flags are really used internally (cannot be set by users), but we expose whether the feature is supported or not in the payment service definition endpoints. Because the credentials verification endpoint is for internal use only (by the dashboard), users probably won't care about whether that feature is supported or not, as there's no endpoint they can use that would be affected by it.

ipeluffo (approved):

PR #134, opened 2022-04-13
Ta 1175 open api add browser info object to api spec

tculigGr4vy:
# Description

https://gr4vy.atlassian.net/browse/TA-1175

## Checklist

- [ X] My code follows the style guidelines of this project
- [ X] I have performed a self-review of my own changes
- [ X] I have run `yarn test` to make sure my changes pass all linters
- [ X] I have pulled the latest changes from the upstream `main` branch

## Contribution guidelines


reference/resources/browser_info.yaml
```diff
@@ -0,0 +1,68 @@
...
+    description: Time-zone offset in minutes between UTC and buyer location.
+    example: 60
+
+  user_device:
+    type: string
+    description: The platform that is being used to access the website.
+    example: "DESKTOP"
+    enum:
+      - "DESKTOP"
+      - "MOBILE"
```
cbetta, line 68:
```suggestion
    example: "desktop"
    enum:
      - "desktop"
      - "mobile"
```

We generally use lower case enums in our API.

thomas-gr4vy:
@tculigGr4vy as discussed, could we add `user_agent` to the spec?

(Context: Resolving the user agent on the backend means that for API users we will have the HTTP client that they are using, and not the user agent of the buyer's browser. If this can be provided via the API then we can require the real value from the user. For Embed the user agent will be provided from `navigator.userAgent`.)

reference/request-bodies/transaction_create_request.yaml
```diff
@@ -156,3 +156,9 @@ properties:
     example: "123456789012345"
     nullable: true
     default: null
+
+  browser_info:
+    allOf:
```
thomas-gr4vy, line 161: Do we need the `allOf` here?

reference/resources/browser_info.yaml
```diff
@@ -0,0 +1,67 @@
...
+    enum:
+      - "desktop"
+      - "mobile"
+
+  user_agent:
+    type: string
+    description: The user agent string for the current browser.
+    example: |-
+      Mozilla/5.0 (darwin) AppleWebKit/537.36
+      (KHTML, like Gecko) jsdom/16.7.0
```
thomas-gr4vy, line 67: You may optionally provide these additional parameters (they are present in the existing browser info schema):
- `xid` (string)
- `challenge_window_size` (one of `"01"`, `"02"`, `"03"`, `"04"` or `"05"` [default])

tculigGr4vy: Sure, but it wasn't in the ticket so...do we need it?

reference/request-bodies/transaction_create_request.yaml
cbetta: Yes

reference/resources/browser_info.yaml
cbetta: Let's leave it out

cbetta (approved):

reference/resources/browser_info.yaml
thomas-gr4vy, line 67:
```suggestion
      (KHTML, like Gecko) jsdom/16.7.0
  accept_header:
    type: string
    description: The `Accept` header of the request from the buyer's browser.
    example: "*/*"
```
Sorry @tculigGr4vy we also need to add an optional `accept_header` property following a change to TA-1176. We won't pass this a value from Embed, in which case it will default to using the request header.

thomas-gr4vy (changes requested):

cbetta (dismissed):

reference/resources/browser_info.yaml
```diff
@@ -0,0 +1,71 @@
...
+  accept_header:
+    type: string
+    description: The `Accept` header of the request from the buyer's browser.
+    example: "*/*"
```
cbetta, line 71:
```suggestion
    example: "*/*"

```

cbetta (approved):

PR #137, opened 2022-04-21
feat: added payment_method_id to card-details endpoint [TA-1301]

thomas-gr4vy:
# Description

Adds the `payment_method_id` to the query string of the `[GET] /card-details` endpoint, allowing users to query card details using **either** the card's BIN or the ID of a stored payment method.

Updates the response to return the `id` as either the BIN or payment method ID, dependant upon which was provided in the original request.

Re [TA-1301](https://gr4vy.atlassian.net/browse/TA-1301)

## Checklist

## Contribution guidelines


reference/paths/card-details.yaml
```diff
@@ -2,12 +2,15 @@
 operationId: get-card-details
 
 summary: Get card details
-description: Gets information about a card.
+description: |-
+  Gets information about a stored payment card using that
+  cards `payment_method_id` or any other payment card using that cards BIN.
```
cbetta, line 7:
```suggestion
  Retrieves details about a card using either that
  card's BIN or for stored cards the `payment_method_id`.
```

reference/resources/card_details.yaml
```diff
@@ -14,11 +14,13 @@ properties:
...
-    maxLength: 8
-    description: The 6-8 digit BIN of the card.
-    example: "424242"
+    oneOf:
+      - format: bin
+        minLength: 8
+        maxLength: 8
+        description: The 8 digit BIN of the card.
+        example: "42424242"
+      - $ref: "./../resources/payment_method.yaml#/properties/id"
```
cbetta, line 23: Do you think we should just return the BIN even for a stored card? In other words keep this as it was?

cbetta (changes requested):

reference/resources/card_details.yaml
thomas-gr4vy: As discussed in Slack, the `id` will be either the 8-digit BIN (when looking up card details using the BIN), or `null` (when looking up card details using the `payment_method_id`).

danielknell (approved):

PR #171, opened 2022-06-24
Update methods.yaml

douglaseggleton:
# Description

`googlepay_pan_only` is not a valid method in a transaction response - need to confirm that it isn't used in another request.

## Checklist

## Contribution guidelines


gryevns:
I think this is valid - do we need to add `googlepay_pan_only` [here](https://github.com/gr4vy/gr4vy-openapi/blob/d99159f69c9f5133e0e72ee55ba2a75bfae00707/reference/request-bodies/google_pay_request.yaml#L15-L20)?

reference/request-bodies/google_pay_request.yaml
```diff
@@ -18,6 +18,7 @@ properties:
     example: googlepay
     enum:
       - googlepay
+      - googlepay_pan_only
```
gryevns, line 21: Sorry, double checked this and it's not the case, we only need `googlepay` so can revert this line. https://github.com/gr4vy/core-api/blob/429bbff9951d1109a57a488af003e2703dbb54f8/src/api/routers/payment_methods/schemas.py#L268

thomas-gr4vy: As per Gary's comment, `googlepay_pan_only` is not required. It is only used internally to Core API now. 👍

douglaseggleton, line 21:
```suggestion
```

PR #177, opened 2022-07-04
feat: added details to payment method [TA-2017]

thomas-gr4vy:
# Description

- Adds the `details` element to a `PaymentMethod` which is intended to hold method-specific properties.
- Adds `details` for the `card` method which includes: `scheme`, `expiration_date`, `first6` and `last4`.

Re [TA-2017](https://gr4vy.atlassian.net/browse/TA-2017)

## Checklist

## Contribution guidelines


reference/resources/payment_method_details_card.yaml
```diff
@@ -0,0 +1,24 @@
+---
+title: Card
+description: A credit or debit card payment method.
+type: object
+x-tags:
+  - Payment Methods
+
+properties:
+
+  card_type:
```
cbetta, line 10: Should this just be type?

thomas-gr4vy: We use `type` across the API to denote the resource, so I went with `card_type` to avoid confusion, and to be consistent with the `/card-details` endpoint which also uses `card_type` (for the same reason).

cbetta (approved):

PR #185, opened 2022-07-13
Required fields for wallets & TH tax ID kind

cbetta:
Re TA-2146

* Adds `required_fields` to payment options 
* Adds new `PH` tax ID kind
* Renamed `CardRequiredFields` to `RequiredFields` as they are no longer card specific.

---

Example

```json
{
  "items": [
    {
      "type": "payment-option",
      "method": "string",
      "icon_url": "https://cdn.gr4vy.app/card.svg",
      "mode": "string",
      "label": "Pay by Card",
      "can_store_payment_method": true,
      "context": {
        "merchant_name": "string",
        "supported_schemes": [
          "amex"
        ],
        "approval_ui": {
          "height": "300px",
          "width": "300px"
        },
        "required_fields": {
          "first_name": true,
          "last_name": true,
          "email_address": true,
          "phone_number": true,
          "address": {
            "city": true,
            "country": true,
            "postal_code": true,
            "state": true,
            "house_number_or_name": true,
            "line1": true
          },
          "tax_id": true
        }
      },
      "payment_method_tokenization_enabled": true
    }
  ]
}
```


reference/resources/required_fields.yaml
```diff
@@ -1,8 +1,8 @@
 ---
-title: Card Required Fields
```
Dave-Elec, line 2: Is required fields descriptive?

Dave-Elec (changes requested):

reference/resources/required_fields.yaml
cbetta: I think so. There is a description with more details.

Dave-Elec: I meant the file name.

cbetta:
@douglaseggleton what's your thoughts on fitting `required_fields` under context? I have trouble understanding what needs to go in context and what not.

reference/resources/required_fields.yaml
Dave-Elec: Okay.

douglaseggleton:
> IMO `context` should be reserved for payment method specific data (ie. we can't rely on the presence of specific items as they can differ between payment methods). This is how it currently works with Google/Apple Pay; both add additional items to the `context` which are not necessarily present for other payment methods. If `required_fields` is applied across all methods then my suggestion would be to move it to the top level.

100% agree with @thomas-gr4vy 👍🏻

thomas-gr4vy:
> @thomas-gr4vy @douglaseggleton just spotted these comments while we just shipped the core API change that does put this under context. Would this be an easy change to move?

As @douglaseggleton highlighted, `approval_ui` is within `context` too, so I think it's fine to continue as-is and migrate both to the top-level as a future improvement?

PR #202, opened 2022-08-25
feat: payment-services endpoint should return only active or deleted [TA-2401]

Dave-Elec:
# Description

This PR updates /payment-services endpoint to include `deleted` query parameter.
This allows either deleted or active payment services to be listed.

Jira: [TA-2401](https://gr4vy.atlassian.net/browse/TA-2401)

## Checklist

## Contribution guidelines


reference/paths/payment_services__list-payment-services.yaml
```diff
@@ -9,6 +9,7 @@ parameters:
   - $ref: "../query-parameters/limit-100.yaml"
   - $ref: "../query-parameters/cursor.yaml"
   - $ref: "../query-parameters/method.yaml"
+  - $ref: "../query-parameters/list_payment_services__deleted.yaml"
```
cbetta, line 12: Can we rename this to `deleted.yml`?

reference/query-parameters/deleted.yaml
```diff
@@ -0,0 +1,9 @@
+---
+schema:
+  type: boolean
+  example: false
```
thomas-gr4vy, line 4:
```suggestion
  example: true
  default: false
```

reference/query-parameters/deleted.yaml
```diff
@@ -0,0 +1,9 @@
...
+in: query
+name: deleted
+description: |-
+  When set to `true`, filter for payment services that are deleted.
+  When set to `false`, filter for payment services that are active.
```
thomas-gr4vy, line 9:
```suggestion
  Filters the results to only show items which have been deleted.
  By default, deleted items will not be returned.
```

reference/query-parameters/deleted.yaml
thomas-gr4vy: As a user is unlikely to ever use `deleted=false` (it's unnecessary) we should instead show the intended usage of filtering to only show deleted items by settings `deleted=true`.

reference/query-parameters/deleted.yaml
thomas-gr4vy: We made this parameter generic when we renamed it to `deleted` so we should make the description generic also.

thomas-gr4vy (changes requested):

thomas-gr4vy (approved):
Looks good to me, as long as @cbetta is happy too.

cbetta (approved):

PR #205, opened 2022-09-05
feat(card-details): add property for scheme icon url

douglaseggleton:
# Description

Add `scheme_icon_url` to the card details endpoint.

# Questions

Do we want to differentiate consumer icons from dashboard icons by calling this a `badge_url` - potentially this would require a contract change in `payment-options` if we went with this approach.

## Checklist

## Contribution guidelines


cbetta (approved):
I like that the icon URL has a specific name to show it's for the scheme. I like this instead of `icon_url` or even `badge_url`.

PR #221, opened 2022-10-18
chore: sync address resource

gryevns:
# Description

Updates `address` resource to match actual API implementation. This means marking all fields as [optional](https://github.com/gr4vy/core-api/blob/179654ff46e30bf15f9a027937229f183fbc1cba/src/api/routers/buyers/schemas.py#L108-L117).

## Checklist

## Contribution guidelines


reference/resources/address.yaml
```diff
@@ -1,8 +1,7 @@
 ---
 title: Address
 type: object
-description: |-
-  The billing address of the buyer.
+description: The address of the buyer.
```
bloowitt, line 4:
```suggestion
description: An address for the buyer.
```
Not sure, just in the case of shippingAddress which uses the same resource, the buyer can have more than one so "the" doesn't entirely fit.

cbetta (approved):

PR #224, opened 2022-10-20
feat: Add GET /report-executions endpoint [TA-2784]

szeying-lee:
# Description
Changes:
* Documents the GET /report-executions endpoint
* Updates the response schemas for report endpoints

Fixes # (issue)
**TA-2784**

## Checklist

## Contribution guidelines


reference/openapi.yaml
```diff
@@ -730,10 +734,16 @@ components:
       $ref: "./resources/report_execution.yaml"
     ReportExecutionUrl:
       $ref: "./resources/report_execution_url.yaml"
+    ReportExecutionWithReportSummary:
```
federicojasson, line 737: This is how schemas relate to each other now (and this applies to all endpoints, not just the new one):

- `Report` contains a `ReportExecutionSummary` (in the `latest_execution` field).
- `ReportExecution` contains a `ReportSummary` (in the `report` field).

(so, the "full" schemas contain their counterpart summary schema)

The summaries obviously include a subset of the "full" schema fields (and they don't contain other schemas, to avoid circular structures). This means that:

- `Report` equals to `ReportSummary` plus other fields.
- `ReportExecution` equals to `ReportExecutionSummary` plus other fields.

(just mentioning this to give a hint on how `allOf` could be used)

**Note**: perhaps the changes here achieve in the same result, but I'd keep the same schema names as in Core API. That would be: we could update the existing `ReportExecution` (instead of introducing a new `ReportExecutionWithReportSummary` schema, and its plural version), while defining a new `ReportExecutionSummary` (and using it for `latest_execution`).

reference/paths/reports__list-all-report-executions.yaml
```diff
@@ -0,0 +1,32 @@
+---
+operationId: list-all-report-executions
+summary: List all report executions
+description:
+  Returns a list of executions of all reports.
```
federicojasson, line 5: [minor] Maybe we can be a bit more explicit about it: `... a list of executions belonging to any report`?

reference/resources/report_executions_with_report_summaries.yaml
```diff
@@ -0,0 +1,23 @@
+---
+title: Report executions
+type: object
+description: A list of report executions.
+x-tags:
+  - Reports
+
+properties:
+  "items":
```
federicojasson, line 9: [minor] Are quotes required? I've seen other list schemas without them 🤔

reference/openapi.yaml
szeying-lee: Refactored report executions here: https://github.com/gr4vy/gr4vy-openapi/pull/224/commits/562a45159c0f2b09b4a0472dd0c670d44ca57c65
And reports: https://github.com/gr4vy/gr4vy-openapi/pull/224/commits/778f3c09fc17f51db4c75f854c468336c7cf7608

reference/resources/report_executions_with_report_summaries.yaml
szeying-lee: Have removed them: https://github.com/gr4vy/gr4vy-openapi/pull/224/commits/778f3c09fc17f51db4c75f854c468336c7cf7608
I think I added these when I was wrestling with the linter ☹️

reference/paths/reports__list-all-report-executions.yaml
szeying-lee: https://github.com/gr4vy/gr4vy-openapi/pull/224/commits/778f3c09fc17f51db4c75f854c468336c7cf7608 and https://github.com/gr4vy/gr4vy-openapi/pull/224/commits/112dcb7882734b1493d4d03f448b6193d005fc9e

victormferrara (approved):
LGTM!

federicojasson (approved):
LGTM!

PR #234, opened 2022-10-27
feat(reporting): Update Reports APIs with scheduling details [TA-2711]

szeying-lee:
# Description
Changes:
* Adds new filters to list Reports API
* Adds scheduling related properties when configuring a new report
* Adds scheduling related properties to `Report` and `ReportExecutionSummary` resources
* Slightly reword Reports APIs description to indicate possibility of configuring one-off and recurring reports

Fixes # (issue)
[TA-2711](https://gr4vy.atlassian.net/browse/TA-2711)

## Checklist

## Contribution guidelines


reference/query-parameters/list_reports__schedule.yaml
```diff
@@ -0,0 +1,13 @@
...
+  items:
+    type: string
+    enum:
+      - monthly
+      - once
+  example: [once, monthly]
+in: query
+name: schedule
+description: |-
+  Filters for reports that have matching `schedule` values.
```
hasier, line 13: Should this be `Filter` like [here](https://github.com/gr4vy/gr4vy-openapi/pull/234/files#diff-b1cfb3eae728d95adb7d80def95d78857c1bb8c5e29ef27106c6612920bea1a7R8), or should the other one be `Filters`?

reference/request-bodies/report_create_request.yaml
```diff
@@ -26,6 +26,58 @@ properties:
...
+      can only be set to `true` if this is a recurring report.
+
+      If this value is set to `true`, the report will be executed at
+      the `next_execution_at` date and time.
+
+      If this is a recurring report and this value is set to `false`,
+      executions of the report will not occur until this value is set
+      to `true`.
+
+      If this value is set to `null`, `schedule_enabled` will
```
hasier, line 60: [minor] `If this value is not provided, ...`.

Also, do we allow sending `None` in the API? Should we allow skipping the value, but if provided, make the client send an actual value?

reference/request-bodies/report_create_request.yaml
```diff
@@ -26,6 +26,58 @@ properties:
...
+      automatically be set to `false` if `schedule` is `once` and set to
+      `true` otherwise.
+    example: true
+    nullable: true
+
+  schedule_timezone:
+    type: string
+    description: |-
+      The time zone which the report's executions will be scheduled in.
```
hasier, line 69: [minor] `The timezone where the report's executions will be scheduled.`?

reference/request-bodies/report_create_request.yaml
```diff
@@ -26,6 +26,58 @@ properties:
...
+      This value is used to compute the report's `next_execution_at`
+      value and is only relevant when this is a recurring report.
+
+      This value must be set to the time zone's descriptive name as
+      presented in the IANA time zone database. For example, to
+      schedule reports in the time zone of New York, set this value to
+      `US/Eastern`.
+    example: "Europe/London"
+    default: "Etc/UTC"
```
hasier, line 78: If it is not nullable, why is there a `default`?

reference/request-bodies/report_update_request.yaml
```diff
@@ -22,3 +22,17 @@ properties:
...
     nullable: true
+
+  schedule_enabled:
+    type: boolean
+    description: |-
+      Indicates whether the report's scheduling is enabled. This value
+      can only be set to `true` if this is a recurring report.
+
+      When this value is set to `true`, the report will be executed at
+      the `next_execution_at` date and time.
```
hasier, line 33: Is `next_execution_at` something visible and editable to the client? If not, should we explain how it is re-calculated based on the current logic? (I remember some conversation where we mentioned when this is switched, we'll recalculate the period from `now()` or something similar?)

reference/resources/report.yaml
```diff
@@ -22,6 +22,15 @@ properties:
       The date and time this report was last updated.
     example: "2013-07-16T19:23:00.000+00:00"
 
+  next_execution_at:
+    type: string
+    format: date-time
+    description: |-
+      The date and time this report will next be executed. This
+      value is only present if this is a recurring report.
```
hasier, line 30: Is it only present then, or is it `None` for one-offs?

reference/resources/report.yaml
```diff
@@ -30,6 +39,43 @@ properties:
     example: "Transactions that failed to authorize in April 2022"
     nullable: true
 
+  schedule:
+    type: string
+    description: |-
+      Specifies the schedule of this report.
+
+      If this report is a one-off report, this value is always `once`.
```
hasier, line 47: [minor] I wouldn't say `always`, it feels like we imply that when it's not one-off it may change.

reference/resources/report_execution_summary.yaml
```diff
@@ -43,3 +43,23 @@ properties:
...
+        type: string
+        format: date-time
+        description: |-
+          The date and time used by the system as a reference point to
+          compute dynamic parameters.
+        example: "2013-07-16T19:23:00.000+00:00"
+      reference_timezone:
+        type: string
+        description: |-
+          The time zone which the dynamic parameters are computed to.
```
hasier, line 64: [minor] `The timezone used to compute datetime dynamic parameters.`?

reference/request-bodies/report_create_request.yaml
szeying-lee: It's current API behaviour, yes 😅 I think because we set the default depending on the `schedule` value, the funky logic means that you can send in `None` 🤔 https://github.com/gr4vy/core-api/blob/main/src/api/routers/reports/schemas.py#L268

reference/request-bodies/report_create_request.yaml
szeying-lee: It's optional, which means if you don't provide it, it will be populated with `Etc/UTC`, and it's not nullable meaning you can't send in `"schedule_timezone": null`. Since we can specify whether a field is required and whether a field is nullable, I took it to mean these two things.

reference/resources/report.yaml
```diff
@@ -30,6 +39,43 @@ properties:
...
+
+      If this report is a recurring report, this value is set to the
```
federicojasson, line 49: "If this report is a recurring report" -> "If this is a recurring report"

reference/resources/report.yaml
```diff
@@ -22,6 +22,15 @@ properties:
       The date and time this report was last updated.
     example: "2013-07-16T19:23:00.000+00:00"
 
+  next_execution_at:
+    type: string
+    format: date-time
+    description: |-
+      The date and time this report will next be executed. This
```
federicojasson, line 29: We may add "executed, provided that `schedule_enabled` is true" or something like that.

reference/resources/report.yaml
```diff
@@ -30,6 +39,43 @@ properties:
...
+      the `next_execution_at` date and time.
+
+      When this value is set to `false`, future executions of this
+      report are paused until this value is set to `true` again.
+    example: true
+
+  schedule_timezone:
+    type: string
+    description: |-
+      The time zone which the next execution will be scheduled in. This
```
federicojasson, line 74: Should it be "The time zone **in** which" (I may be completely wrong)

reference/resources/report.yaml
```diff
@@ -30,6 +39,43 @@ properties:
...
+      value is used to calculate this report's `next_execution_at` value
+      and is only relevant if this is a recurring report.
```
federicojasson, line 76: In practice, we use the `schedule_timezone` as the execution `context.reference_timezone`, and that value is used for dynamic params. Given that one can use dynamic params for `once` reports (even if it doesn't make a lot of sense), then this statement wouldn't be truthy.

I guess the question is: should we expose these internal details here? 🤔

reference/request-bodies/report_create_request.yaml
```diff
@@ -26,6 +26,58 @@ properties:
...
+    nullable: true
+
+  schedule_timezone:
+    type: string
+    description: |-
+      The time zone which the report's executions will be scheduled in.
+      This value is used to compute the report's `next_execution_at`
+      value and is only relevant when this is a recurring report.
+
+      This value must be set to the time zone's descriptive name as
```
federicojasson, line 73: I wonder if "descriptive name" may be misinterpreted. For example, `Etc/GMT-3` is a valid value, but ain't very "descriptive".

reference/request-bodies/report_update_request.yaml
federicojasson: ☝️ Yeah, I think it makes sense to explain it. Here's the logic:

```python
        if command.new_values.get("schedule_enabled") is True:
            if report.schedule == types.ReportSchedule.once:
                raise errors.ValidationError(
                    message="One-off reports can't be scheduled",
                    location=["schedule_enabled"],
                    type="value_error",
                )

            if (
                not report.schedule_enabled
                and typing.cast(datetime.datetime, report.next_execution_at) < now
            ):
                # Because next_execution_at is a timestamp
                # in the past, it needs to be recalculated
                report.next_execution_at = reports.get_next_execution_at(
                    base_timestamp=now,
                    schedule=report.schedule,
                    timezone=report.schedule_timezone,
                )
```

The human-friendly explanation:

If the scheduling is being enabled **after being disabled**, then `next_execution_at` is updated **if and only if** its current value is a past datetime. The`next_execution_at` value is then set to the next closest datetime in the future (which depends on `schedule` and `schedule_timezone`).

reference/request-bodies/report_create_request.yaml
```diff
@@ -26,6 +26,58 @@ properties:
...
+      presented in the IANA time zone database. For example, to
+      schedule reports in the time zone of New York, set this value to
+      `US/Eastern`.
```
federicojasson, line 76: I'd use `America/New_York` as an example here, as this is the canonical value.

Not sure if this is something we want to explicitly mention in the documentation, but there are canonical values and aliases. Canonicals are the "primary, preferred zone name". Both types would work, but it's probably best if we use canonical values in the examples.

victormferrara (commented):
Not much to add, I think everything's very well covered with current comments 👍

reference/request-bodies/report_create_request.yaml
ipeluffo, line 69: We probably should mention that this is also used to filter data for scheduled reports.

Then, for example, a monthly report with timezone `America/Argentina` will have monthly data for that timezone which will be different than a month of data for UTC.

cc: @federicojasson

reference/resources/report.yaml
ipeluffo: > Given that one can use dynamic params for once reports (even if it doesn't make a lot of sense), then this statement wouldn't be truthy.

Although it's possible to send dynamic params for one-off reports, I feel like it's confusing and I'm sort of thinking if we should not allow users to combine them because for a one-off report a user can just set the how date range they need.

So, even if we add the validation above or not and we may be lying, I feel like leaving the documentation as it's doesn't sound bad.

> In practice, we use the schedule_timezone as the execution context.reference_timezone, and that value is used for dynamic params.

Agree, we shouldn't expose this as it's purely internal logic

reference/request-bodies/report_create_request.yaml
ipeluffo: Marking as resolved to discuss in the other thread

reference/resources/report.yaml
ipeluffo: Discussed offline with @federicojasson :
1. Leave the comment that this is relevant for recurring reports
2. Add some explanation that this affects the calculated date range when using datetimes place holders, i.e. dynamic parameters.

Can be something like:
> This timezone is also used to calculate the date range for reports that use placeholders for dynamic timestamps.

reference/query-parameters/list_reports__schedule.yaml
szeying-lee: Fixed: https://github.com/gr4vy/gr4vy-openapi/pull/234/commits/482335b2490149ae6a0a0a5bfaa532c1614fef4c

reference/request-bodies/report_create_request.yaml
szeying-lee: Reworded here: https://github.com/gr4vy/gr4vy-openapi/pull/234/commits/692b816c46fe804895e55fe5e35d551bd7def5ed#diff-dac36ed726e612f841c5ce10499164bef474fafa1b5a276bbd5c47be63ae6eaaR60

reference/request-bodies/report_create_request.yaml
szeying-lee: I went with Fede's suggestion in the end: https://github.com/gr4vy/gr4vy-openapi/pull/234#discussion_r1007217783
Updated here: https://github.com/gr4vy/gr4vy-openapi/pull/234/commits/692b816c46fe804895e55fe5e35d551bd7def5ed#diff-dac36ed726e612f841c5ce10499164bef474fafa1b5a276bbd5c47be63ae6eaaR69
And here: https://github.com/gr4vy/gr4vy-openapi/pull/234/commits/3e0facbbcaca229cf18c17390c98d6eef467b6a2#diff-8618b4d69869ffc13a95f92c5575ccef0d4affe1a6948b5b688ca7865e791591R75

reference/request-bodies/report_update_request.yaml
szeying-lee: Added that description here: https://github.com/gr4vy/gr4vy-openapi/pull/234/commits/a92159922efd83730672688d65cf43e2db0832d2
(and a missing word: https://github.com/gr4vy/gr4vy-openapi/pull/234/commits/01dd94eecf939c2fae57b5aebecf64ed9ca1d118)

reference/resources/report.yaml
szeying-lee: Maybe this? https://github.com/gr4vy/gr4vy-openapi/pull/234/commits/3e0facbbcaca229cf18c17390c98d6eef467b6a2#diff-8618b4d69869ffc13a95f92c5575ccef0d4affe1a6948b5b688ca7865e791591R30

reference/resources/report.yaml
szeying-lee: Removed `always`: https://github.com/gr4vy/gr4vy-openapi/pull/234/commits/3e0facbbcaca229cf18c17390c98d6eef467b6a2#diff-8618b4d69869ffc13a95f92c5575ccef0d4affe1a6948b5b688ca7865e791591R48

reference/resources/report_execution_summary.yaml
szeying-lee: I reworded most of the `context` properties, hopefully this sounds better: https://github.com/gr4vy/gr4vy-openapi/pull/234/commits/8c8142fe08646b8ad03bdfcd4195c36961c3ecf8#

reference/resources/report.yaml
szeying-lee: Done: https://github.com/gr4vy/gr4vy-openapi/pull/234/commits/3e0facbbcaca229cf18c17390c98d6eef467b6a2#diff-8618b4d69869ffc13a95f92c5575ccef0d4affe1a6948b5b688ca7865e791591R48

reference/resources/report.yaml
szeying-lee: Done: https://github.com/gr4vy/gr4vy-openapi/pull/234/commits/3e0facbbcaca229cf18c17390c98d6eef467b6a2#diff-8618b4d69869ffc13a95f92c5575ccef0d4affe1a6948b5b688ca7865e791591R29

reference/resources/report.yaml
szeying-lee: Sounds better IMO, changed here: https://github.com/gr4vy/gr4vy-openapi/commit/3e0facbbcaca229cf18c17390c98d6eef467b6a2#diff-8618b4d69869ffc13a95f92c5575ccef0d4affe1a6948b5b688ca7865e791591R75
 and https://github.com/gr4vy/gr4vy-openapi/pull/234/commits/692b816c46fe804895e55fe5e35d551bd7def5ed#diff-dac36ed726e612f841c5ce10499164bef474fafa1b5a276bbd5c47be63ae6eaaR69

reference/resources/report.yaml
szeying-lee: Added here: https://github.com/gr4vy/gr4vy-openapi/commit/692b816c46fe804895e55fe5e35d551bd7def5ed#diff-dac36ed726e612f841c5ce10499164bef474fafa1b5a276bbd5c47be63ae6eaaR71
And here: https://github.com/gr4vy/gr4vy-openapi/commit/3e0facbbcaca229cf18c17390c98d6eef467b6a2#diff-8618b4d69869ffc13a95f92c5575ccef0d4affe1a6948b5b688ca7865e791591R77

reference/request-bodies/report_create_request.yaml
szeying-lee: Removed `descriptive`: https://github.com/gr4vy/gr4vy-openapi/commit/692b816c46fe804895e55fe5e35d551bd7def5ed#diff-dac36ed726e612f841c5ce10499164bef474fafa1b5a276bbd5c47be63ae6eaaR75

reference/request-bodies/report_create_request.yaml
szeying-lee: Changed the example: https://github.com/gr4vy/gr4vy-openapi/commit/692b816c46fe804895e55fe5e35d551bd7def5ed#diff-dac36ed726e612f841c5ce10499164bef474fafa1b5a276bbd5c47be63ae6eaaR75

reference/request-bodies/report_create_request.yaml
```diff
@@ -26,6 +26,59 @@ properties:
...
+    example: true
+    nullable: true
+
+  schedule_timezone:
+    type: string
+    description: |-
+      The time zone in which the report's executions will be scheduled.
+      This value is used to compute the report's `next_execution_at`
+      value and is only relevant when this is a recurring report. This
+      time zone is also used to calculate the date range for reports
```
federicojasson, line 72: "date range" --> "timestamp range"?

reference/resources/report.yaml
```diff
@@ -30,6 +40,45 @@ properties:
...
+      report are paused until this value is set to `true` again.
+    example: true
+
+  schedule_timezone:
+    type: string
+    description: |-
+      The time zone in which the next execution will be scheduled. This
+      value is used to calculate this report's `next_execution_at` value
+      and is only relevant if this is a recurring report. This time zone
+      is also used to calculate the date range for reports that use
```
federicojasson, line 78: "date range" --> "timestamp range"?

reference/resources/report.yaml
```diff
@@ -30,6 +40,45 @@ properties:
...
+      date-time placeholders as dynamic timestamps.
```
federicojasson, line 79: Is `as dynamic timestamps` redundant/unnecessary in this sentence?

reference/request-bodies/report_create_request.yaml
```diff
@@ -26,6 +26,59 @@ properties:
...
+      that use date-time placeholders as dynamic timestamps.
```
federicojasson, line 73: Is `as dynamic timestamps` redundant/unnecessary in this sentence?

federicojasson (commented):
Just two similar pair of comments.

reference/request-bodies/report_create_request.yaml
szeying-lee: Done: https://github.com/gr4vy/gr4vy-openapi/pull/234/commits/cd0e36b2ab628ae67116ed90f9a856b935a2a73f

reference/resources/report.yaml
szeying-lee: https://github.com/gr4vy/gr4vy-openapi/pull/234/commits/cd0e36b2ab628ae67116ed90f9a856b935a2a73f

reference/resources/report.yaml
szeying-lee: I think so. I reworded it a bit with a short description of what `date-time placeholders` are like in some of the other descriptions, hopefully this is better: https://github.com/gr4vy/gr4vy-openapi/pull/234/commits/cd0e36b2ab628ae67116ed90f9a856b935a2a73f

reference/request-bodies/report_create_request.yaml
szeying-lee: https://github.com/gr4vy/gr4vy-openapi/pull/234/commits/cd0e36b2ab628ae67116ed90f9a856b935a2a73f

federicojasson (approved):

PR #269, opened 2022-12-20
feat(flow): add post endpoint to list payment options

Moosapoor:
# Description

Adds new post `/payment-options` endpoint. 

Fixes # (issue)
[TA-3398](https://gr4vy.atlassian.net/browse/TA-3398)

## Checklist

## Contribution guidelines

[TA-3398]: https://gr4vy.atlassian.net/browse/TA-3398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TA-3398]: https://gr4vy.atlassian.net/browse/TA-3398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Moosapoor:
> I have no idea. Those were there and I had nothing to do with those parts

bloowitt:
We should fix them but not in this PR :D

bloowitt (dismissed):
Can not be merged until the changes to `core-api` have been merged.

reference/paths/payment_options__post-list-payment-options.yaml
```diff
@@ -0,0 +1,42 @@
...
+      Returns a list of available payment options for the given query
+      parameters.
+    content:
+      application/json:
+        schema:
+          $ref: "../openapi.yaml#/components/schemas/PaymentOptions"
+
+  "400":
+    description: |-
+      Returns an error if  any of the query parameters are not recognised.
```
gryevns, line 31: There are no query parameters on the POST endpoint & extra space between "if" & "any"

reference/request-bodies/payment_options_request.yaml
```diff
@@ -0,0 +1,75 @@
...
+      the value specified for `country` in `payment_method` will be assumed
+      implicitly.
+
+  metadata:
+    description: |-
+      Any additional information about the transaction that you would like to
+      store as key-value pairs. This data is passed to payment service
+      providers that support it. Please visit https://gr4vy.com/docs/ under
+      `Connections` for more information on how specific providers support
+      metadata.
```
gryevns, line 61: Maybe link to the connections page if possible? Otherwise I don't think "Connections" needs to be rendered as code.

reference/request-bodies/payment_options_request.yaml
```diff
@@ -0,0 +1,75 @@
...
+    description: |
+      A supported ISO-4217 currency code.
+
+      For redirect requests, this value must match the one specified for
+      `currency` in `payment_method`.
+
+  country:
+    type: string
+    example: US
+    nullable: true
```
gryevns, line 44: Is this nullable?

gryevns (dismissed):

reference/request-bodies/payment_options_request.yaml
Moosapoor: Yes I double checked. All parameters are nullable except `locale`

gryevns: In the existing endpoint we have a default value for `locale`: https://github.com/gr4vy/core-api/blob/5ada2623f391a524d1bcd46c4dd76af95510fed9/src/api/routers/payment_options/endpoints.py#L24 - I am not sure `null` is a valid choice? Also the example of `US` might be invalid?

reference/request-bodies/payment_options_request.yaml
```diff
@@ -33,32 +27,22 @@ properties:
     type: string
     example: USD
     description: |
-      A supported ISO-4217 currency code.
-
-      For redirect requests, this value must match the one specified for
-      `currency` in `payment_method`.
```
gryevns: What was the reason for removing this? I think there is some value in specifying the ISO type.

reference/request-bodies/payment_options_request.yaml
```diff
@@ -9,16 +9,10 @@ x-tags:
 properties:
   amount:
     description: |-
-      The monetary amount to create an authorization for, in the smallest
-      currency unit for the given currency, for example `1299` cents to create
-      an authorization for `$12.99`.
-
-      If the `intent` is set to `capture`, an amount greater than zero must
-      be supplied.
```
gryevns: I think this should be reverted too - let's document the values rather than the usage (i.e. flow).

reference/request-bodies/payment_options_request.yaml
Moosapoor: So I think I should add a default value for `locale` in `post` method as well. 
Also the example of `US` is for `country` so I think it's valid

reference/request-bodies/payment_options_request.yaml
Moosapoor: I just tried to make the descriptions same as the `get` method. I can revert them by the way

reference/paths/payment_options__post-list-payment-options.yaml
```diff
@@ -0,0 +1,43 @@
...
+      parameters.
+    content:
+      application/json:
+        schema:
+          $ref: "../openapi.yaml#/components/schemas/PaymentOptions"
+
+  "400":
+    description: |-
+      Returns an error if the requested JSON could not be parsed or any of the
+      JSON keys are not recognised.
```
gryevns, line 32:
```suggestion
      Returns an error if the request was badly formatted or missing required
      fields.
```

reference/request-bodies/payment_options_request.yaml
gryevns: If they are all nullable then do we need to document this?

Moosapoor:
127 0 0 1_8080_#tag_Payment-Options_operation_post-list-payment-options

reference/request-bodies/payment_options_request.yaml
```diff
@@ -0,0 +1,70 @@
...
+      If the `intent` is set to `capture`, an amount greater than zero must
+      be supplied.
+    type: integer
+    example: 1299
+
+  locale:
+    type: string
+    example: en-US
+    pattern: ^[a-z]{2}(?:-[A-Z]{2})?$
+    default: en
```
bloowitt, line 26: I've noticed the old endpoint is documented as `default: en-US` but the default is actually `en` as documented here. Would be nice to update the old endpoint while touching this.

(Same with metadata, old endpoint has it documented as string instead of object)

Moosapoor: Sure. I'll change the old endpoint as well

Moosapoor: As the old endpoint's parameters are coming from query string, is it correct to change the `metadata` type to object instead of string?

bloowitt: Let me check how we do it with other endpoints.

Moosapoor: I already did that. Let me know if I should revert it

Moosapoor: All the endpoints use the same yaml file if they're getting the `metadata` within query params

bloowitt: The OpenAPI 3.0 spec supports objects as query parameters in a much richer way than we are using: https://www.baeldung.com/openapi-json-query-parameters , but it's not worth it probably

bloowitt (approved):

PR #275, opened 2023-01-04
feat(3DS): 3DS error response schema

bloowitt:
# Description
Jira: TA-3558

Please include a summary of the change and which issue is fixed.
Screenshot 2023-01-05 at 11 42 43

If you are a Gr4vy employee, please also reference the related Zoho tickets.
If you do not have a related Zoho ticket and this is more than a bug fix, then
create one now.

Fixes # (issue)
Re ... (Zoho ticket)

## Checklist

## Contribution guidelines

[TA-3558]: https://gr4vy.atlassian.net/browse/TA-3558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


reference/resources/three_d_secure_error.yaml
```diff
@@ -0,0 +1,35 @@
...
+    type: string
+    description: Detail for the error.
+    example: "The threeDSCompInd must be 'Y' when successful"
+    maxLength: 2048
+
+  code:
+    type: string
+    description: |-
+      The error code. See
+       https://docs.3dsecure.io/3dsv2/specification_220.html#error-object.
```
gryevns: Probably best not to include third party docs (we shouldn't expose the underlying service publicly and means something else to update if we swap the implementation in core-api).

bloowitt: removed it just before the comment (https://github.com/gr4vy/gr4vy-openapi/pull/275/commits/c80074f921ac6ee1ec8f07a81211b19c5011ad3b)

gryevns:
I think this will require more changes, e.g. `method` should be optional - we should revisit this once we understand how and what data structures we're saving.

reference/resources/three_d_secure_error.yaml
```diff
@@ -0,0 +1,43 @@
...
+title: 3-D Secure Error Data
+type: object
+description: |-
+  Details about the error resulting from 3DS processing a Transaction.
+
+x-tags:
+  - Transactions
+
+required:
+  - description
```
bloowitt, line 11: should we make these all not required? from discussion it seems we might get into situations where we have some but not all of them

gryevns: From reviewing the core-api change it looks like the keys will always be returned given an error response exists so required makes sense. That being said I think you could remove the `required` section completely because that is the default value?

bloowitt: The `three_d_secure_data_v1` and `v2` contain the `required` section even though their fields are all also required - just did a quick check and if I remove `required` from here they are rendered as not required. 
Screenshot 2023-01-11 at 14 19 32
I think for homogeneity we should keep this `required` or remove it from the others too.

reference/resources/three_d_secure_error.yaml
```diff
@@ -0,0 +1,43 @@
...
+    maxLength: 3
+    minLength: 3
+
+  component:
+    type: string
+    minLength: 1
+    maxLength: 1
+    description: |-
+      Code indicating the 3-D Secure component that identified the error..
+    example: "C"
```
gryevns, line 43: I think these should all be marked nullable

gryevns (approved):

PR #276, opened 2023-01-09
TA-3593 Define open_loop for PaymentService

hasier:
# Description

TA-3593

Depends on https://github.com/gr4vy/core-api/pull/2045

## Checklist

## Contribution guidelines


reference/request-bodies/payment_service_update_request.yaml
```diff
@@ -172,6 +172,20 @@ properties:
...
       enabled if the payment service definition supports it).
 
+  open_loop:
+    type: boolean
+    example: true
+    nullable: true
+    description: |-
+      Defines if this service works as an open-loop service.
+
+      If this value is not provided or is set to `null`, it will be set
```
federicojasson, line 182: For an update, the behaviour is the following:

- If `open_loop` is not set, the value isn't changed (because we iterate using `for key in document.__fields_set__`).
- If `open_loop` is `None`, we set the value of the PSD.

For the creation, `None` and "not present" works in the same way.

This is the same for `network_tokens_enabled`, it may be worthwhile to correct it in this PR as well.

hasier: Hmm... This is tricky because we use [this same definition yaml for both creation and update](https://github.com/gr4vy/gr4vy-openapi/blob/a07a0c12289281dd1ed431143f9a53e59d208a5c/reference/request-bodies/payment_service_create_request.yaml#L17), adding wording for both sounds like it might get confusing? Also, creating yet another separate definition sounds somewhat error prone, so maybe wording for all cases is not that bad... 😕

reference/request-bodies/payment_service_update_request.yaml
```diff
@@ -172,6 +172,20 @@ properties:
...
+      to the value of `open_loop` in the payment service definition.
+
+      If `open_loop_toggle` is `false`, `open_loop` should either not be
+      provided or set to `null`, or it will fail with a validation error.
+      This will then be set to the value of `open_loop`.
```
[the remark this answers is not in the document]
hasier: True, it's explained in the paragraph above. Done! https://github.com/gr4vy/gr4vy-openapi/pull/276/commits/9b47d48e1878d3474e90193df1533ab8084e157f

reference/request-bodies/payment_service_update_request.yaml
federicojasson: Well, we already have one file for creation and one for the update, we could have a third one with the common fields. As you see fit.

hasier: I have seen that with other definitions we have the values duplicated in both create and update (e.g. buyers), so I've replicated the same thing. I have also slightly reworded all the flags, hopefully I made them more consistent https://github.com/gr4vy/gr4vy-openapi/pull/276/commits/50fa766a03667e24b77b26251a21df7be78236d6

federicojasson (approved):

PR #280, opened 2023-01-10
feat(merchant-profile): add merchant profile to payment service response

bloowitt:
# Description

TA-3618 ( PR 1 of 3)
Screenshot 2023-01-10 at 12 55 25

## Checklist

## Contribution guidelines


sach-gr4vy (dismissed):

reference/resources/merchant_profile.yaml
```diff
@@ -0,0 +1,33 @@
+---
+title: Merchant Profile
+type: object
+description: Merchant profile for the different card schemes.
+
+properties:
+  amex:
+    type: object
+    nullable: true
```
gryevns, line 9: Will these be nullable or will we create a merchant profile row for each card scheme when a new payment service is created?

reference/resources/merchant_profile_scheme.yaml
```diff
@@ -0,0 +1,38 @@
...
+    description: |-
+      URL to send when calling 3DS through this scheme.
+
+  merchant_acquirer_id:
+    type: string
+    nullable: true
+    description: |-
+      Merchant ID to use when calling 3DS through this scheme.
+
+  country_code:
```
gryevns, line 26: For consistency?
```suggestion
  merchant_country_code:
```

reference/resources/merchant_profile_scheme.yaml
```diff
@@ -0,0 +1,38 @@
...
+    type: string
+    nullable: true
+    minLength: 3
+    maxLength: 3
+    description: Acquirer bin to use when calling 3DS through this scheme.
+
+  category_code:
```
gryevns, line 33:
```suggestion
  merchant_category_code:
```

reference/resources/payment_service.yaml
```diff
@@ -151,6 +151,16 @@ properties:
     maxLength: 4
     description: Merchant category code that describes the business.
 
+  merchant_profile:
+    type: object
+    nullable: true
+    description: |-
+      An object containing a key for each supported card scheme (Amex,
+      Discover, Mastercard and Visa), and for each key an object with
+      the merchant profile for this service and the corresponding scheme.
```
gryevns, line 160: Suggest removing the card scheme names from the description, it should be obvious from the nested schema what values are returned and saves us from forgetting to update this if we change them.

reference/resources/merchant_profile.yaml
bloowitt: I'm using the same resource (MerchantProfile) for the update, hence it being nullable (for them to send it as null to delete the scheme). This is in line with what we discussed around it when decomposing, but I'm open to changing that (or forcing it to not be nullable on response but be nullable on update)

gryevns: As we're adding a new table to store merchant profiles I think it would be nice to have all columns as not nullable (this matches the form validation we currently have on Dashboard).

bloowitt: After conversation today, made them non-nullable (kept each scheme in `merchant_profile` as nullable as one could not configure a specific scheme)

gryevns (approved):

gryevns (approved):

PR #294, opened 2023-02-07
chore: Update `payment_method.external_identifier` descriptions in `TransactionRequest` [TA-3706]

szeying-lee:
# Description
New validation on `payment_method.external_identifier` has been added in `TransactionRequest`: https://github.com/gr4vy/core-api/pull/2264

This PR updates the description for the `store` property in `TransactionRequest` following these changes.

Fixes # (issue)
[TA-3706](https://gr4vy.atlassian.net/browse/TA-3706)

## Checklist

## Contribution guidelines

[TA-3706]: https://gr4vy.atlassian.net/browse/TA-3706?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


victormferrara (dismissed):

reference/request-bodies/transaction_create_request.yaml
```diff
@@ -72,6 +72,9 @@ properties:
       * The flag has to be set to `true` when the `payment_source` is set to
       `recurring` or `installment`, and `merchant_initiated` is set to `false`.
 
+      * The flag has to be set to `true` if setting an `external_identifier`
+      for the `payment_method`.
+
```
hasier, line 77: Shouldn't we word it the other way around? I mean, `store` is more meaningful than `external_identifier`, so maybe we should put more emphasis on providing `external_identifier` if `store=true` instead? That way we can also discourage its use if `store=false`. wdyt? @szeying-lee @victormferrara

victormferrara: It would be more like providing `external_identifier` **only** if `store=true`, right? Sounds good to me too.

szeying-lee: I've moved the description here: https://github.com/gr4vy/gr4vy-openapi/pull/294/commits/bae986cc38cfa8083eccd534959c74ae24d0c71a
Looks a bit messy as I had to override the descriptions for `external_identifier` as they were also being used for the [store payment method endpoint](https://github.com/gr4vy/gr4vy-openapi/blob/main/reference/paths/payment_methods__store-payment-method.yaml).

reference/request-bodies/transaction_create_request.yaml
```diff
@@ -50,12 +50,45 @@ properties:
...
+            example: card-323444
+            nullable: true
+      - type: object
+        allOf:
+          - $ref: "../openapi.yaml#/components/schemas/RedirectRequest"
+        properties:
+          external_identifier:
+            type: string
+            description:
+              An external identifier that can be used to match the account
```
szeying-lee, line 72: I copied this over from the `RedirectRequest` schema, but should this actually be
```suggestion
              An external identifier that can be used to match the card
```
?

victormferrara: The redirect payment method does not necessarily reference a card underneath, so I guess that this is referring to PayPal accounts and so on.

victormferrara (approved):

PR #296, opened 2023-02-13
fix: CreateTransaction schema's payment_method object titles

szeying-lee:
# Description
Changes made in https://github.com/gr4vy/gr4vy-openapi/pull/294 introduced a bug:
image
Above, the titles of overridden `payment_method` objects were replaced with `object` which was not the case when running the server locally with `yarn run start`. This PR tries to fix this issue.

Any advice on how run the open API spec locally so that it matches production would be appreciated.

## Checklist

## Contribution guidelines


victormferrara (approved):
Sounds like a fair attempt to solve the issue 👍

I've been trying to reproduce it locally but failed 😞

I don't know much about `yarn`, but seeing the [defined targets](https://github.com/gr4vy/gr4vy-openapi/blob/883b2314648927a6a970b1c3119888b893e43e08/package.json#L6) I though that `yarn build:public-spec` and `yarn start:public` could help, but I ended up with the same result.

The GitHub action in the end calls a Netlify endpoint, which I don't know what it is either. I didn't find anything in our repos so I guess it's an external service. Maybe it has some custom config and we could replicate it when calling [preview-docs](https://redocly.com/docs/cli/commands/preview-docs/) 🤷. I'm afraid I only have very random guesses xD

hasier:
Maybe @cbetta can help us better understand the setup, and maybe a better way to reproduce this?

szeying-lee:
@victormferrara @hasier Thanks both! I have let @cbetta know, but will merge the PR for now 🤞.

PR #299, opened 2023-02-13
chore: Improve wording for card payment method redirect_url (TA-2495)

hasier:
# Description

Jira: TA-2495

Improve the wording of the `redirect_url` field when creating transactions to make it clearer when it's needed.

## Checklist

## Contribution guidelines


reference/request-bodies/tokenized_request.yaml
```diff
@@ -30,12 +30,15 @@ properties:
...
-      We strongly recommended providing a `redirect_url` for
-      stored cards when 3-D Secure is enabled and `three_d_secure_data`
-      is not provided. This will be appended with both a
-      transaction ID and status (e.g. `https://example.com/callback?
-      gr4vy_transaction_id=123&gr4vy_transaction_status=capture_succeeded`)
-      after 3-D Secure has completed.
+      This value is mandatory for stored redirect payment methods.
+      For stored cards, we strongly recommend providing a `redirect_url`
+      either when 3-D Secure is enabled and `three_d_secure_data`
+      is not provided, or when using connections where 3DS is enabled.
```
victormferrara, line 36: Is "connection" a term that we usually use here? At least at top-level we are using "payment service", right? Or is it not the same in this case?

reference/request-bodies/transaction_create_request.yaml
```diff
@@ -62,6 +62,20 @@ properties:
                   flag is set to `true`.
                 example: card-323444
                 nullable: true
+              redirect_url:
```
victormferrara, line 65: This override is what I've seen with @szeying-lee that is not working fine when rendering the final [documentation page](https://docs.gr4vy.com/reference). I've been checking this kind of usages and mostly they are not working, although a few of them do work fine 🤷

reference/request-bodies/tokenized_request.yaml
hasier: To be honest, we have a bit of a mishmash of connector, payment service and PSP all around the docs 😅  Happy to use the one you feel is more consistent in this case.

reference/request-bodies/transaction_create_request.yaml
federicojasson: This is due to this script: https://github.com/gr4vy/gr4vy-openapi/blob/main/scripts/build-sdk-spec.js#L8

Is replacing the whole `payment_method` object by `{'$ref': `#/components/schemas/TransactionPaymentMethodRequest`}` if I'm correct.

federicojasson (approved):

reference/request-bodies/transaction_create_request.yaml
victormferrara: Hmmmm I'd say that's only executed in the `sdk` workflow to generate the spec that later on is converted into the different API clients, and in this case we're dealing with the `docs` workflow.

PR #377, opened 2023-07-26
feat(account-updater): Document last_replaced_at and has_replacement fields [TA-4951]

federicojasson:
# Description

Jira: TA-4951

This PR documents the new payment method fields related with the Account Updater feature:
- `last_replaced_at`
- `has_replacement`

⚠️ **Do not merge until the feature is released.**

## Checklist

## Contribution guidelines


reference/resources/payment_method--tokenized.yaml
```diff
@@ -98,3 +98,20 @@ properties:
       be used for. If this value is `null` the payment method may be
       used in multiple countries.
     nullable: true
+
+  last_replaced_at:
+    type: string
+    nullable: true
+    description: |-
+      The date and time when this payment method was last replaced.
+      This value is always `null` for non-card payment methods.
```
cbetta, line 107: @federicojasson I think it would make sense to explain what would cause a replacement here. E.g. an account updater.

Also, why is this not called `last_updated_at`?

reference/resources/payment_method--tokenized.yaml
```diff
@@ -98,3 +98,20 @@ properties:
...
+    format: date-time
+    example: "2023-07-26T19:23:00.000+00:00"
+
+  has_replacement:
+    type: boolean
+    nullable: false
+    description: |-
+      Whether this payment method has a pending replacement that hasn't been applied yet.
+      This value is always `false` for non-card payment methods.
```
cbetta, line 116:
```suggestion
      Whether this card has a pending replacement that hasn't been applied yet.
      This value is always `false` for non-card payment methods.
```

If it's always a card then let's call it a card.

Again, I think this needs a bit more explanation as to what a "replacement" is. E.g. a new set of card details.

cbetta (changes requested):
Left some comments. Also not sure about the `replace` terminology over `update`.

cbetta:
@federicojasson also, when/where are we documenting the endpoints to request an update?

reference/resources/payment_method--tokenized.yaml
federicojasson: Sure, I can add some more details. Eventually we may also want a section in the [docs](https://github.com/gr4vy/gr4vy-docs-mintlify) that explains in detail what Account Updater is and how it works.

federicojasson:
@cbetta

> Also not sure about the `replace` terminology over `update`.

I'd say that "update" is overloaded in this context, since we have an `updated_at` field, which refers to any change, not just to card details. WDYT?

> when/where are we documenting the endpoints to request an update?

This will be documented in TA-4823 (PR coming later today).

cbetta:
> @cbetta
> 
> > Also not sure about the `replace` terminology over `update`.
> 
> I'd say that "update" is overloaded in this context, since we have an `updated_at` field, which refers to any change, not just to card details. WDYT?
> 
> > when/where are we documenting the endpoints to request an update?
> 
> This will be documented in TA-4823 (PR coming later today).
> 

I agree. Yet this feature will be sold as an Account Updated. Not a Card Replacer. I wonder if it makes sense to stay closer to the language a merchant would expect.

jadk89:
I think it's ok to leave it as "last_replaced_at" and "has_replacement", reason for these two fields is we might have an updated card while the old card is not yet expired, we will try with the old card until it gets declined or expires. We then have a replacement for the card available. For any other type of update we indeed have the `updated_at` field.

reference/resources/payment_method--tokenized.yaml
federicojasson: Done: [Extend descriptions of new fields](https://github.com/gr4vy/gr4vy-openapi/pull/377/commits/eaf2761ecc766374fd277474ed635b71e1c94c36)

reference/resources/payment_method--tokenized.yaml
federicojasson: Done: [Extend descriptions of new fields](https://github.com/gr4vy/gr4vy-openapi/pull/377/commits/eaf2761ecc766374fd277474ed635b71e1c94c36)

jadk89 (approved):

cbetta (approved):

PR #378, opened 2023-07-27
feat(account-updater): Document loon_client_key and loon_secret_key fields [TA-4950]

federicojasson:
# Description

Jira: TA-4950

This PR documents the new fields used to configure Loon API authentication:
- `loon_client_key`
- `loon_secret_key`

Endpoints affected by this change:
- `POST /merchant-accounts`
- `PUT /merchant-accounts/{id}`

⚠️ **Do not merge until the feature is released.**

## Checklist

## Contribution guidelines


cbetta (approved):

jadk89 (approved):

federicojasson (commented):
I've written documentation for a new field we're adding in [TA-4835](https://gr4vy.atlassian.net/browse/TA-4835): [Document loon_enabled field](https://github.com/gr4vy/gr4vy-openapi/pull/378/commits/a8633470d2c0dfbd41fb62e30e23f81944fcb21d)

reference/resources/merchant_account.yaml
```diff
@@ -81,6 +81,13 @@ properties:
     example: e50fa0da-903d-4d54-aacc-4cac57d48df2
     nullable: true
 
+  loon_enabled:
+    type: boolean
+    description: |-
+      Whether Loon API keys have been set up for this merchant account.
+      Loon is the Account Updater service used by Gr4vy.
+    example: false
+
```
cbetta, line 90: @Moosapoor most of our API tries to be symmetrical. E.g. what goes in comes out. Why do we not return the loon client and secret key?

cbetta (changes requested):
I'd like the API to be symmetrical

reference/resources/merchant_account.yaml
cbetta: cc @federicojasson

federicojasson: There's a technical reason for it. The Loon API keys are stored in the Secret Manager. That means that, in order for us to return them (asterisks `*****` actually), we would have to make a call to the Secret Manager. This isn't a huge issue for `GET /merchant-accounts/{id}`, but for `GET /merchant-accounts` we would have to make one call per record.

federicojasson: A workaround: return `loon_client_key` and `loon_secret_key` fields with asterisks if `loon_enabled = TRUE`, or `None` if `loon_enabled = FALSE`. Would that make more sense? @cbetta

federicojasson: Done: [Replace loon_enabled by loon_*_key fields](https://github.com/gr4vy/gr4vy-openapi/pull/378/commits/163dd11a2b1cff2db4238307bad976535b6328fd)

cbetta: You read my mind. 

Also, is loon_client_key realy a secret?

cbetta (approved):

reference/resources/merchant_account.yaml
federicojasson: > Also, is loon_client_key realy a secret?

It may not be strictly a secret (in the same way a username is not sensitive information but we still want to protect it), but given it's a 32 hexadecimal character string, I'm not sure if exposing it in the API adds any value.

Additionally, it simplifies things technically, as we can store both keys in the same secret record.

PR #390, opened 2023-08-09
feat(3DS): store decline data

gryevns:
# Description

TA-5050

Extend 3DSv2 transaction response data with optional `transaction_reason`

## Checklist

## Contribution guidelines


reference/resources/three_d_secure_v2.yaml
```diff
@@ -0,0 +1,79 @@
...
+allOf:
+  - required:
+      - directory_transaction_id
+
+    properties:
+      cavv:
+        type: string
+        description: The cardholder authentication value or AAV.
+        example: "3q2+78r+ur7erb7vyv66vv8="
+        nullable: true
```
gryevns, line 14: @cbetta before several of these properties were not optional (cavv/eci etc.) so this is still a breaking change to support an optional transaction reason - are we happy with this approach?

cbetta: @gryevns yeah I think so. I think only 1 merchant is using this and they have asked for this change. We will document it.

cbetta (approved):

PR #401, opened 2023-09-06
fix: Remove checkout session fields API endpoint from SDKs

steve-gr4vy:
Screenshot 2023-09-06 at 17 06 36


cbetta:
@steve-gr4vy I am not sure what this does. 

What does that Java code do? Which API does this point to?

`PUT /checkout/sessions/:id`?

steve-gr4vy:
> @steve-gr4vy I am not sure what this does.
> 
> What does that Java code do? Which API does this point to?
> 
> `PUT /checkout/sessions/:id`?

It's referenced here: https://github.com/gr4vy/gr4vy-openapi/blob/main/reference/request-bodies/checkout_sessions_fields_update_request.yaml

And used here: https://github.com/gr4vy/gr4vy-openapi/blob/main/reference/paths/checkout_sessions_fields__update-fields.yaml

cbetta:
So I think this API should not be public actually. This is what secure fields calls to store the PCI data on the TX. Do you think we should have this in the SDK? I am happy to mark this as internal.

steve-gr4vy:
> So I think this API should not be public actually. This is what secure fields calls to store the PCI data on the TX. Do you think we should have this in the SDK? I am happy to mark this as internal.

There is a bunch of stuff generated in the SDK that I don't think should be.. agree this isn't needed

cbetta:
@steve-gr4vy shall I remove this one?

cbetta:
@steve-gr4vy updated. I will merge when checks pass.

cbetta (approved):

cbetta:
@steve-gr4vy can I get a 👍🏻 seeing as this is now basically my PR.

PR #409, opened 2023-10-03
feat: Add gift card processing and vaulting APIs

cbetta:


reference/paths/gift_card_service_definitions__get-gift-card-service-definition.yaml
cbetta, line 1: [Info] Not really related but fixes an incorrectly named file from a previous PR.

reference/paths/payment_methods__list-buyer-gift-cards.yaml
```diff
@@ -0,0 +1,36 @@
+---
+operationId: list-buyer-gift-cards
+summary: List gift cards buyer
+description: |-
+  Returns a list of all stored, non-empty gift cards and
+  their balances for a buyer in a summarized format.
```
cbetta, line 6: [Info] Rather than adding a filter, this API always returns what we need in Embed: gift cards, including their balance, filtering our zero value gift cards.

reference/request-bodies/gift_card_balance_request.yaml
```diff
@@ -0,0 +1,32 @@
+---
+title: Gift Card Balance (Request)
+type: object
+description: Check the balance for a stored/new gift card.
+
+x-tags:
+  - Request Bodies
+
```
cbetta, line 8: [Info] This is a schema used by the SDK only. It combined the 2 different request schemas for new/stored cards into 1.

reference/request-bodies/transaction_gift_card_request.yaml
```diff
@@ -0,0 +1,48 @@
...
+title: Transaction gift card (Request)
+type: object
+description: Create a transaction with this gift card.
+
+x-tags:
+  - Request Bodies
+
+required:
+  - amount
+
```
cbetta, line 11: [Info] This is a schema used by the SDK only. It combined the 2 different request schemas for new/stored cards into 1.

reference/request-bodies/transaction_refund_request.yaml
```diff
@@ -13,12 +13,31 @@ properties:
...
+      - payment-method
+      - gift-card
+
+  target_id:
+    type: string
+    nullable: true
+    description: |-
+      The optional ID of the instrument to refund for. This is only required when
+      the `target_type` is set to `gift-card`.
+    example: c23ea83f-1b1c-4584-a0e8-78ef8c041949
```
cbetta, line 43: [Info] This allows a refund to target a gift card instead of the main payment method. This is only needed to target a gift card refund.

reference/request-bodies/transaction_refund_request.yaml
```diff
@@ -13,12 +13,31 @@ properties:
...
+    type: string
+    description: |-
+      The target type to refund for. This can be used to target a gift card
+      to refund to instead of the main payment method.
+    example: gift-card
+    default: payment-method
+    nullable: true
+    enum:
+      - payment-method
+      - gift-card
```
cbetta, line 35: [Info] I decided not to expose gift card redemptions in the API. Instead, the refund happens on a transaction and targets the gift card by ID.

reference/resources/gift_card--snapshot.yaml
```diff
@@ -0,0 +1,74 @@
...
+  - Gift Cards
+
+properties:
+  type:
+    type: string
+    description: The type of this resource.
+    example: gift-card
+    enum:
+      - gift-card
+
```
cbetta, line 16: [Info] Although this is listed as a gift-card snapshot, this object kind of behaves as a merger of a `gift-card` and a redemption. In my original proposal I a redemption linked under this resource, but in the end this was simpler.

reference/resources/gift_card--snapshot.yaml
```diff
@@ -0,0 +1,74 @@
...
+      than `amount`.
+    example: 1299
+    minimum: 1
+    maximum: 99999999
+
+  currency:
+    type: string
+    example: USD
+    description: |-
+      The ISO-4217 currency code that this gift card was charged for.
```
cbetta, line 74: [Info] These fields are technically part of the redemption on this gift card for this TX. Not sure if we should move this out or not into its own object.

reference/paths/gift_cards__check-gift-card-balances.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+responses:
+  "201":
+    description: |-
+      Returns the balances for each of the gift cards. Balances are returned in the same
+      order they are requested.
+    content:
+      application/json:
+        schema:
+          $ref: "../openapi.yaml#/components/schemas/GiftCards--Summary"
```
cbetta, line 23: [Info] Rather than returning balance objects, this simply returns the same object as in `GET /buyers/gift-cards` with an updated `balance`.

reference/resources/transaction.yaml
```diff
@@ -60,6 +60,16 @@ properties:
...
+  completed:
+    type: boolean
+    description: |-
+      A single status indicator to help a merchant understand if every
+      payment method and/or gift cards have reached the desired `intent`.
+
+      In short, when all payment method and gift cards have reached at
+      least the `succeeded`/`*_succeeded` state for the given intent then
+      the transaction has completed.
+
```
cbetta, line 72: [Info] This is a request from Wpay. They wanted an object a merchant could observe to know if the entire transaction had completed.

reference/resources/transaction.yaml
```diff
@@ -73,26 +83,38 @@ properties:
...
+    type: integer
+    description: |-
+      The amount for this transaction that has been authorized for the `payment_method`.
+      This can be less than the `amount` if gift cards were used.
+
+      This can also be more than the actual `captured_amount`, and part of this
+      amount may have been refunded.
     example: 1299
     minimum: 0
     maximum: 99999999
```
cbetta, line 102: [Info] This is a new field that represents the actual amount charged to the `payment_method`, removing any amounts charged to gift cards from `amount`. In case of an `intent=capture` this would be set to the `capture_amount`.

reference/resources/gift_card.yaml
```diff
@@ -0,0 +1,78 @@
+---
+title: A gift card
+type: object
+description: A gift card stored for a buyer, or used in a transaction.
+
+x-tags:
+  - Gift Cards
+
```
cbetta, line 8: [Info] This is the regular gift card object. It's lacking the redemption and balance fields seen in the Summary/Snapshot versions.

reference/request-bodies/gift_card_balances_request.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+properties:
+  items:
+    type: array
+    description: One or more gift cards to check balances for.
+    items:
+      x-model-name: GiftCardBalanceRequest
+      oneOf:
+        - $ref: "../openapi.yaml#/components/schemas/GiftCardBalanceNewRequest"
+        - $ref: "../openapi.yaml#/components/schemas/GiftCardBalanceStoredRequest"
```
cbetta, line 19: [Question] Does it make sense for this API to also be usable with stored cards? The API would become a lot simpler without it, and stored cards can already be queried through `GET /buyers/gift-cards`?

reference/paths/refunds__refund-all.yaml
```diff
@@ -0,0 +1,44 @@
+---
+operationId: refund-all
+
+summary: Refund all instruments in a transaction
+description: |-
+  Refunds a transaction fully across all instruments.
+
+tags:
+  - Transactions
```
cbetta, line 9: [Info] This is a new API that tries to refund all methods on a transaction.

victormferrara: 1. Would it make sense use the current endpoint with `target_type=all`?
2. Are we allowing partial refunds in this endpoint?

cbetta: 1. We could, but it would change the response type (1 object vs an array of objects) which makes the SDKs a pain to work with. Hence, keeping a predictable response type.
2. No, there's no amount passed to this endpoint.

victormferrara: Oh yeah, I missed the different returned schema type 👍

reference/resources/gift_card--snapshot.yaml
```diff
@@ -0,0 +1,74 @@
...
+  refunded_amount:
+    type: integer
+    description: |-
+      The amount refunded for this gift card. This can not be larger
+      than `amount`.
+    example: 1299
+    minimum: 1
+    maximum: 99999999
+
+  currency:
```
victormferrara, line 70: Needed? If this type is nested in the transaction it should be the same currency that the main transaction object, right?

reference/resources/gift_card.yaml
```diff
@@ -0,0 +1,78 @@
...
+properties:
```
victormferrara, line 9: Should we add the buyer if linked?

reference/paths/gift_cards__check-gift-card-balances.yaml
```diff
@@ -0,0 +1,53 @@
...
+  - Gift Cards
+
+requestBody:
+  content:
+    application/json:
+      schema:
+        $ref: "../openapi.yaml#/components/schemas/GiftCardBalancesRequest"
+
+responses:
+  "201":
```
victormferrara, line 16: 200 status code? There's not really any object being created.

reference/resources/gift_card--summary.yaml
```diff
@@ -0,0 +1,81 @@
...
+
+  id:
+    type: string
+    format: uuid
+    nullable: true
+    description: |-
+      The ID of this gift card. This may be `null` if the gift card
+      is not a stored gift card.
+    example: e6cdf979-87e2-4796-8ff6-9784d5aed893
+
```
victormferrara, line 25: `merchant_account_id`?

reference/resources/gift_card--summary.yaml
```diff
@@ -0,0 +1,81 @@
...
+    description: |-
+      The 3 digits after the `bin` of the full gift card number.
+    example: "554"
+
+  last4:
+    type: string
+    description: |-
+      The last 4 digits for the gift card.
+    example: "1234"
+
```
victormferrara, line 43: No `expiration_date`?

reference/resources/gift_card--summary.yaml
```diff
@@ -0,0 +1,81 @@
...
+  balance_updated_at:
+    type: string
+    description: |-
+      The date and time when this gift card balance was last fetched.
+    format: date-time
+    nullable: true
+    example: "2013-07-16T19:23:00.000+00:00"
+
+  balance_error_code:
```
victormferrara, line 67: Do we have a set of these errors to map or do we just return what the provider gives us?

reference/resources/transaction.yaml
victormferrara: Only for succeeded states or for all final ones? (failures, declines, etc).

cbetta: @victormferrara I think the goal for the merchant is to know if the entire TX has succeeded or not. The alternative is a new kind of status that's either `succeeded` (all payments worked), `pending` (we're not sure yet) or `failed` (one or more of the payments failed/declined).

reference/resources/gift_card--snapshot.yaml
cbetta: True, I can remove.

reference/resources/gift_card.yaml
cbetta: Good point, can do.

reference/paths/gift_cards__check-gift-card-balances.yaml
cbetta: Correct, I can update

reference/resources/gift_card--summary.yaml
cbetta: Not sure it has much use, but we do the same for the payment method summary (--Tokenized) so will do it here too.

reference/request-bodies/gift_card_balances_request.yaml
victormferrara: I think it does make sense for completeness sake.

reference/request-bodies/transaction_gift_card_request.yaml
victormferrara: Not really, right? This is used as-is in the transactions create type now.

reference/resources/transaction.yaml
victormferrara: Hmmm ok, I'm concerned about some kind of polling until this flag is `true` and that never happening.

reference/paths/gift_cards__get-gift-card.yaml
```diff
@@ -0,0 +1,36 @@
...
+      application/json:
+        schema:
+          $ref: "../openapi.yaml#/components/schemas/Error404NotFound"
+
+  default:
+    description: Returns a generic error.
+    content:
+      application/json:
+        schema:
+          $ref: "../openapi.yaml#/components/schemas/ErrorGeneric"
```
hasier, line 36: Have we used this before? When would this be returned?

reference/paths/gift_cards__check-gift-card-balances.yaml
hasier: Does this mean that a regular `GET /gift-cards` should also return balances?

reference/request-bodies/transaction_gift_card_request.yaml
hasier: Not sure I follow this conversation... We are going to need the 2 separate schemas in order to accept both stored and non-stored Gift Cards. Would this break that?

reference/request-bodies/transaction_refund_request.yaml
hasier: Any particular reason not to do it separately? It felt simpler when we discussed it as the response of the endpoint would be more consistent and specific.

reference/resources/gift_card--snapshot.yaml
```diff
@@ -0,0 +1,74 @@
...
+      The first 6 digits of the full gift card number.
+    example: "412345"
+
+  sub_bin:
+    type: string
+    description: |-
+      The 3 digits after the `bin` of the full gift card number.
+    example: "554"
+
+  last4:
```
hasier, line 37: I guess `label` was a more generic name to group both card and non-card PM display info then?

reference/resources/gift_card--snapshot.yaml
```diff
@@ -0,0 +1,74 @@
...
+    type: string
+    description: |-
+      The last 4 digits for the gift card.
+    example: "1234"
+
+  expiration_date:
```
hasier, line 43: [info] In QC [this is defined as a full datetime](https://plinazcertenv.sharepoint.com/:w:/r/sites/Qwikcilver-AU-External/Shared%20Documents/Gr4vy/Integration%20Documents/%5BExternal%5D%20Gr4vy%20-%20Interface%20Control%20Document.docx?d=wa28282fa2ee346a0af6ebe0fb13c90e1&csf=1&web=1&e=joS3Pf&nav=eyJoIjoiNDI4NzExNjE3In0%3D). However, they mention they only take into account the date bit, we probably need to clarify how they round it to better reflect the info.

reference/resources/gift_card--snapshot.yaml
hasier: If we are following the shape of your original proposal, I guess we should move them down to the redemption?
```json
  "gift-cards": [{
    "type": "gift-card",
    "id": "5c3ba1d8-f418-416f-9493-5abf23e0d05d",
    "label": "8334",
    "redemption": {
      "type": "gift-card-redemption",
      "id": "cb6093ad-0166-419c-9378-8d4496b8bf09",
      "status": "succeeded",
      "amount": 500,
      "currency": "USD",
      ...
    }
  }
```

reference/resources/gift_card--summary.yaml
hasier: Description says `raw error code`. If so, maybe we should rephrase this to allow for a Gr4vy-unified error code in the future? Something like `raw_balance_error_code` or `provider_balance_error_code`? Same for `balance_error_message` below.

reference/resources/gift_cards--summary.yaml
```diff
@@ -0,0 +1,15 @@
+---
+title: Gift Cards (Summary)
+type: object
+description: |-
+  A list of gift cards in a short format including the latest balance.
+
+x-tags:
+  - Gift Cards
+
+properties:
```
hasier, line 10: No limits?

reference/resources/transaction.yaml
hasier: Agree with Victor. Also, the definition sounds a bit vague, if the transaction is authorised and we redeem GCs, this becomes `True`, but if then is voided, becomes `False`? It feels to me like each merchant may expect different things here, and it is not a difficult value to calculate as the response will always contain all the info. Personally I'd rather leave this kind of calculation to the clients to handle as they prefer, it's checking 1 txn field + N in the GC array.

reference/resources/transaction.yaml
```diff
@@ -73,26 +83,38 @@ properties:
...
+    maximum: 99999999
+
+  authorized_amount:
+    type: integer
+    description: |-
+      The amount for this transaction that has been authorized for the `payment_method`.
+      This can be less than the `amount` if gift cards were used.
+
+      This can also be more than the actual `captured_amount`, and part of this
+      amount may have been refunded.
```
hasier, line 99: I find this wording a bit difficult to follow... What we are trying to say is that it could happen that there's less amount captured than the initially authorised, right?

reference/resources/transaction.yaml
```diff
@@ -73,26 +83,38 @@ properties:
...
     minimum: 0
     maximum: 99999999
 
   refunded_amount:
     type: integer
     description: |-
-      The refunded amount for this transaction. This can be the total or a
-      portion of the captured amount.
+      The amount for this transaction that has been refunded for the `payment_method`.
+      This can be the total or a  portion of the `captured_amount`.
```
hasier, line 117: [typo] Additional space.

reference/resources/gift_card--summary.yaml
cbetta: Good point, adding.

reference/resources/gift_card--summary.yaml
cbetta: That's one for you and @hasier. I prefer a mapped code, but not sure how easy that is. If we can map then I'd prefer to have `balance_error_code` be a mapped value, and `balance_message` is the raw message received.

reference/resources/gift_card--summary.yaml
```diff
@@ -0,0 +1,81 @@
...
+  balance:
+    type: integer
+    description: |-
+      The amount remaining on the balance for this gift card according to the
+      gift card service.
+    example: 1299
+    minimum: 0
+    maximum: 99999999
+
+  balance_updated_at:
```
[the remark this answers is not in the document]
cbetta: Ah yes this is a leftover. Removing.

reference/paths/gift_cards__get-gift-card.yaml
cbetta: @hasier this is our standard error response format with little detail. It's a capture-all for any explicit errors we missed. It's mostly an API requirement.

reference/resources/gift_card--snapshot.yaml
cbetta: @hasier exactly. If you want I can rename this to label as well.

reference/resources/transaction.yaml
cbetta: Fixing

reference/paths/gift_cards__check-gift-card-balances.yaml
cbetta: @hasier so yeah that is the odd one here. Basically, `GiftCard--Summary` has a subset of the fields from `GiftCard` plus the balance fields. We could return the last balance we received with the `GiftCard` as well, but a) it's another piece of data we need to store, and b) its likely out of date at this point. We could simply return the `balance_*` fields for a `GiftCard` but make them null?

reference/request-bodies/transaction_gift_card_request.yaml
cbetta: Yes, you are right. I will fix this as 2 schemas. This schema will become the one used for SDKs

victormferrara: > Not sure I follow this conversation... We are going to need the 2 separate schemas in order to accept both stored and non-stored Gift Cards. Would this break that?

@hasier Consider the [GiftCardBalanceRequest](https://github.com/gr4vy/gr4vy-openapi/blob/b29cecfbb1a737e018d5a616c81f1859de2720aa/reference/request-bodies/gift_card_balances_request.yaml) case. We're defining a union type, but in most languages unions just don't exist, so we need a [type aggregating all the fields](https://github.com/gr4vy/gr4vy-openapi/blob/b29cecfbb1a737e018d5a616c81f1859de2720aa/reference/request-bodies/gift_card_balance_request.yaml) to be able to send ones of the others.

reference/request-bodies/transaction_refund_request.yaml
cbetta: @hasier this was a request from Wpay. In their mind, all refunds are on the TX. Also, in the original example we had redemptions exposed as an object, and the refunds happened on the redemption. This API does away with that and only returns `gift-cards`. So either we add a target to the refund endpoint, or we add a new endpoint to refund a gift card. The problem is that a gift card refund only makes sense within the context of a TX (which we have here) or a redemption (which we no longer expose via the API).

reference/resources/gift_card--snapshot.yaml
cbetta: @hasier does this mean the format would more likely be a `date` (not `date-time`, and not this weird string?)

reference/resources/gift_card--summary.yaml
victormferrara: > `balance_error_code` be a mapped value, and `balance_error_message` is the raw message received

I think this could work, we can map a couple of usual errors leaving a generic one (like `invalid_card_pin_combination` and `error`) and form a message with what we receive from the provider.

reference/resources/gift_card--snapshot.yaml
cbetta: @hasier so I did away with the redemption in this proposal. We could still have them internal but I did not share them external. I could re-introduce it, but it somehow did not make sense to be nested under the gift card cause we dont return it anywhere else.

The alternative approaches I had considered were these 2. Would love your thoughts. 

* Option 1 - Redemptions with nested gift cards
  * transaction
    * gift_card_redemptions
      * gift_card
* Option 2 - Redemptions and gift cards under transactions
  * transaction
    * gift_cards
    * gift_card_redemptions 
      * gift_card_id  

In the end, all of these felt over-engineered and I felt like the merchant didn't need to know about the individual redemptions.

reference/paths/gift_cards__check-gift-card-balances.yaml
hasier: Not particularly a fan of nullable fields for this kind of cases, but also true that a "balance object" as a subset of the GC feels odd, not sure...

reference/resources/gift_card--summary.yaml
hasier: Ok fair, the latest QC docs have some error codes, so we can add some mapping.

@cbetta when you say "`balance_error_message ` is the raw message received", do you mean the actual human-readable message, or the raw code from the provider? In either case, I would still name it something that makes it more straightforward to understand it is not an error we crafted, but something the provider gave us.

reference/resources/gift_cards--summary.yaml
cbetta: @hasier Let's discuss the limit on the list of stored cards separately. I think that's an endpoint validation issue, not a limit on the schema.

reference/resources/transaction.yaml
cbetta: Yes, to be honest that's more of a concern for the `captured_amount` field. I will change the copy.

reference/resources/gift_card--snapshot.yaml
hasier: I thought we used `last4` for card PMs but I have not found the case, so maybe `label` is more consistent with other endpoints. But GCs are very specific and are not wrapping multiple behaviours anyway, so `last4` might be adequate. No strong preferences, just thinking out loud.

